PR #14, opened 2023-05-29
Added k8s api

nirkal:


nirkal:
@tianjunz Is this in the expected format?

tianjunz:
Thanks for your contribution!🚀  I have been able to process the file and get this new updated API. Could you review it and let me know if that is correct or not?

nirkal:
> Thanks for your contribution!🚀 I have been able to process the file and get this new updated API. Could you review it and let me know if that is correct or not?

this looks about right. I've got a naive question (I am still learning how this works). I had included the complete API spec of Kubernetes, after the processing, there are two APIs. I wonder how it is possible to process the complete API spec of k8s.

tianjunz:
It is because the two questions you wrote is related to these two API calls and we just extract that from the entire JSON dataset. If you could come up with some question and api_call pairs, that would be super helpful! Thanks again for contributing!

nirkal:
okay, i start to understand this better now. I can come up with additional question and API call pairs. I actually want to make an exhaustive list, so we are able to fully generate k8s APIs from Gorilla (for all resources, and all actions). This might take some effort.
I have some further questions, would really appreciate your answers.

1. Is the an automatic CICD pipeline planned/setup so that when we push commits to the master, we are able to deploy the changes? I.e when I add the k8s api, how soon the model is able to server the new changes?

2. When a question & api call pair is given how is the extraction happening from json? Is there an automation involved here or its manual effort? In the commit, I see that the extraction for first question/API return python code, but for the second questions, it uses kubectl cli command. How to make this consistent.

tianjunz:
1. Sorry that we are planning to go through the reviewing process to make sure everything looks ok on our end. It is also aimed to provide a high-quality dataset. We will also provide a website to make this task easier! We plan to start community-contributed API Zoo training this week and have the first model released on 06/05. 
2. Yeah, I am not super familiar with Kubernetes. I tried to extract the from the documentation for the second command. The only thing I can find is the cli command. It would be super helpful if the document contains a consistent code template. Please let us know what you think, and we can better assist you in this case.

Alikbra (approved):

53jk1 (approved):

data/apizoo/nirkal.json
```diff
@@ -1,11 +1,30 @@
 [
   {
-    "user_name": "example_username_url",
-    "api_name": "Kubernetes API",
-    "url": "https://raw.githubusercontent.com/kubernetes/kubernetes/release-1.27/api/openapi-spec/swagger.json",
+    "user_name": "nirkal", 
+    "api_name": "Kubernetes Pod List", 
+    "api_call": "io.k8s.api.core.v1.PodList()", 
+    "api_version": "1.0",
```
nirkal, line 6: The api_version should correspond to the K8S API version right? In this case, 1.27.

nirkal:
@tianjunz thanks for the clarification.
If I may ask you , what is the impact if entire JSON file is processed without questions.

@Alikbra @53jk1 What is your opinion? Is it possible to come up with questions coverage vast range of API calls in k8s. Also curious how you plan to use k8s api calls from Gorilla.

Issue #18, opened 2023-05-29
inference/gorilla_eval.py funcion not implemented


eisneim:
in file `inference/gorilla_eval.py` when set  `--device mps ` the following function is not implemented
```
replace_llama_attn_with_non_inplace_operations()
```

ShishirPatil:
Reference PR #19

PR #19, opened 2023-05-29
remove `replace_llama_attn_with_non_inplace_operations` as we don't supported it yet

tianjunz:
Ref: #18 
We haven't added support for Apple's MPS for inference. Contributions welcome!


ShishirPatil (approved):
LGTM

PR #23, opened 2023-05-30
Fix: gorilla_cli.py `model` to `model-path`

tianjunz:
In gorilla_cli.py the argument must be `model-path` and not `model`


ShishirPatil (approved):
LGTM.

PR #33, opened 2023-06-02
Update ast_eval_th.py

mzamini92:
The parsing of API calls into AST trees is now performed in parallel using the `ThreadPoolExecutor` from the `concurrent.futures` module, allowing for faster processing of the API database.
modified the mistypos.


ShishirPatil (approved):
LGTM. Thank you @mzamini92 for parallelizing, and fixing our misspells and linting ❤️

PR #38, opened 2023-06-06
🟢 Releasing Apache 2.0 models MPT/Falcon

ShishirPatil:
In this PR we release the weights for Gorilla 0-shot model for Hugging Face Hub APIs with MPT-7B and Falcon-7B base models 🎉

It chooses from 925 HF APIs in a 0-shot fashion (without any retrieval). We do not filter, nor carry out any post processing either to the prompt nor response. You can play around with Gorilla either through our hosted colab (OpenAI Chat completion API compliant) or you can download the weights and run it locally.

As always, thank you for all the comments and suggestions so far! Keep them coming!!!

🚀 🚀 🚀


tianjunz (approved):
LGTM

Issue #58, opened 2023-07-07
The bm25 and gpt-index scripts ?


zhilizju:
For the different retrievers, we use bm25 (https://en.wikipedia.org/wiki/Okapi_BM25), gpt-index simply uses `Davinci v1` from OpenAI to embed all the documents and do simple cosine similarity match during the inference time. For oracle, we just provided the golden truth answer to Gorilla. Hope this helps and let me know if there are any further questions!

_Originally posted by @tianjunz in https://github.com/ShishirPatil/gorilla/issues/21#issuecomment-1567800181_

Would you be willing to release the bm25 and gpt-index scripts to help the community reproduce the experimental results?

ShishirPatil:
Hey @zhilizju thanks for raising this. What exactly are you referring to? Like how to build and use a retriever?

ShishirPatil:
Thank you for your kind words and 893 out of 904 is a good match :) But yeah, you should get a 100% match - I think others have been able to reproduce it. For BM25, which variant of BM25 are you using? We used  Okapi BM25 (`BM25Okapi` from `rank_bm25`). So we got the embedding for each of the documents (API's in our case) listed [here](https://github.com/ShishirPatil/gorilla/tree/main/data/api) and then did a cosine similarity match. Similarly, even for openai's embeddings, we didn't use the text-search, instead got the embeddings and compared for each query. Can you try using `text-embedding-ada-002-v2` to get the embedding and then do a simple top-1 cosine similarity search? Let me know how it goes or if you run into any issues.

zhilizju:
Great! Thank you very much! It seems that some parts of the [code](https://github.com/ShishirPatil/gorilla/pull/61) were missing, so I added them and reproduced the experiment. However, I still found some differences in the results. For BM25, out of 904, there are 63 differences, and for GPT-index, out of 904, there are 99 differences.

zhilizju:
Of course, but I have been busy with the rebuttal and paper submission lately. I will submit a PR (pull request) after a while. Thank you once again.

ShishirPatil:
Thanks @zhilizju and good luck with the submissions :) Will close this for now!

Issue #32, opened 2023-06-02
[Apibench] Needs to be easy to query for which API and version fully supported


yosun:
Needs to be easy to query for which API and version fully supported

PR #61, opened 2023-07-18
Adding BM25 and GPT retrievers

ShishirPatil:
Gorilla introduces Retriever Aware Training (RAT). Here we include the code for retrievers. Note that while you can use  {prompt + retrieved API} with the 0-shot trained Gorilla LLMs for inference, for optimal performance, we recommend you use {prompt + retrieved API} with RAT trained Gorilla LLM for inference. 

More precisely, BM25 refers to `BM25Okapi` and GPT refers to `text-embedding-ada-002-v2`

Directly relevant to #58 and might be useful for #32 

+ Removing unused legacy code from `ast_eval_*.py`
+ Updating Apache 2.0 license for all of Gorilla code


tianjunz (approved):
LGTM

PR #84, opened 2023-08-06
Add support for Azure, OpenAI, Palm, Anthropic, Cohere, Replicate Models - using litellm

ishaan-jaff:
This PR adds support for models from all the above mentioned providers using litellm https://github.com/BerriAI/litellm

TLDR: gorilla gets:
- more models out of the box, everything litellm integrates with
- instant integration of new models by modifying the `model` param (Easy to add models in the future, just change the model param)

Here's a sample of how it's used:
```
from litellm import completion

## set ENV variables
# ENV variables can be set in .env file, too. Example in .env.example
os.environ["OPENAI_API_KEY"] = "openai key"
os.environ["COHERE_API_KEY"] = "cohere key"

messages = [{ "content": "Hello, how are you?","role": "user"}]

# openai call
response = completion(model="gpt-3.5-turbo", messages=messages)

# cohere call
response = completion("command-nightly", messages)

# anthropic call
response = completion(model="claude-instant-1", messages=messages)
```

### DIFF
--- a/eval/get_llm_responses.py
+++ b/eval/get_llm_responses.py
@@ -17,8 +17,11 @@ import sys
 import json
 import openai
 import anthropic
+import litellm
+from litellm import completion
 import multiprocessing as mp
 import time
+import os
 
 def encode_question(question, api_name):
     """Encode multiple prompt instructions into a single string."""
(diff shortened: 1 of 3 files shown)


ishaan-jaff:
@ShishirPatil can you please take a look at this ? Happy to add docs/tests if this initial commit looks good

eval/get_llm_responses.py
```diff
@@ -52,24 +55,15 @@ def get_response(get_response_input, api_key):
...
             openai.api_key = api_key
-            responses = openai.ChatCompletion.create(
+            responses = completion(
                 model=model,
                 messages=question,
                 n=1,
                 temperature=0,
             )
             response = responses['choices'][0]['message']['content']
-        elif "claude" in model:
```
ishaan-jaff, line 64: litellm manages model I/O for your model choice

ShishirPatil:
Hello @ishaan-jaff, thank you for your PR! It's impressive to see a wrapper that can accommodate multiple models. However, I'm a bit concerned about introducing an additional dependency that might not be essential for standard use cases. While I see its potential value, I'm contemplating whether it's best to include it in a separate release or integrate it into the main releases. I'm flexible on this. Could you initiate a discussion thread titled "Request for Comments - LiteLLM Integration" so we can gather community feedback? Based on that I can review this appropriately. Though at a high level, it looks good, since you basically count on the user to create the `message`. 
Thoughts?

ishaan-jaff:
**Thanks for the feedback @ShishirPatil** 
- started a discussion here: https://github.com/ShishirPatil/gorilla/discussions/89
- as for adding additional dependencies, we can address that by only requiring dependencies when they are used

eg. only install litellm `anthropic` if a user is trying to use claude-2

krrishdholakia:
@ShishirPatil @rajveer43 happy to help as well :)

PR #109, opened 2023-08-22
Update README.md

abhi-databricks:


ShishirPatil (approved):
Thanks for the catch and fix!

PR #115, opened 2023-08-24
Fix use_wandb in ast eval, responses file deletion, wandb artifacts renaming

morganmcg1:
- Fix `use_wandb` to be `args.use_wandb`
- deletes existing responses file if one is found (otherwise data is appended)
- Fix W&B artifacts logging
- Only append to table if line from jsonl is not none
- Enforce artifacts naming conventions (will error if `:` are present)


ShishirPatil (approved):
LGTM

PR #133, opened 2023-10-16
Implemented Blog Index Feature 

NithikYekollu:
Implemented blog indexes for subtitles within blog for first three blogs. Implemented drop down feature to navigate to other blogs in blog index.


ShishirPatil (approved):
LGTM

PR #163, opened 2024-01-30
Added Gmail api to api zoo

saikolasani:
Added the Gmail API (Python) to API Zoo. This includes all of the methods available to users except for the methods that require the client-side encryption (CSE) configuration. This is not accessible by the regular API user.


ShishirPatil (approved):
LGTM. Thanks for the PR @saikolasani

PR #179, opened 2024-02-09
Adding 9 Google Services to API Zoo

meenakshi-mittal:
Exhaustively added python client list for Google Calendar, Chat, Docs, Drive, Search Console, Sheets, Slides, Translate, and Youtube to API Zoo


ShishirPatil:
Hey @meenakshi-mittal thanks for PR! Quick suggestions, 

- For each of the services included, can you flag if all the APIs were exhaustively included? 
- Maybe also include the source page for each of the APIs so we can keep track of changes to API if and when they evolve. 
- Also are you double check that all the APIs are infact just 'v1' and have no further sub-version?

PR #215, opened 2024-02-27
Berkeley Function Calling Leaderboard evaluation scripts and OpenFunctions v2 inference 

ShishirPatil:
Berkeley Function Calling Leaderboard evaluation scripts and OpenFunctions v2 inference


tianjunz (approved):
LGTM

PR #234, opened 2024-03-04
support Gemini model

vandyxiaowei:

PR #237, opened 2024-03-05
Leaderboard update: mistral-large-2402, gemini-1.0-pro and gemma 7-b; Update REST eval. 

HuanzhiMao:
Update leaderboard data and wagon wheel data to reflect new models and the updated rest API eval result.

Change leaderboard to include `mistral-large-2402`, `gemini-1.0-pro`, and `gemma-7b`.
Change leaderboard evals to be consistent with #234

### DIFF
--- a/index.js
+++ b/index.js
@@ -142,7 +142,7 @@ const data = {
     datasets: [
         {
             label: 'GPT-4-0125',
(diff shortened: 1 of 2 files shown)


ShishirPatil (approved):
LGTM

PR #239, opened 2024-03-05
Update openfunctions-v2 chatting format in README.md

tianjunz:
Update openfunctions-v2 chatting format in README.md

### DIFF
--- a/openfunctions/README.md
+++ b/openfunctions/README.md
@@ -165,10 +165,11 @@ def get_prompt(user_query: str, functions: list = []) -> str:
     Returns:
     - str: The formatted conversation prompt.
     """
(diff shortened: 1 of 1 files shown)


ShishirPatil (approved):
Note:
    The openfunctions-v1 template also works, though for "relevance detection" the new v2 format in this PR is preferred.

Use the below code for backward compatibility with openfunctions-v0 and openfunctions-v1 (and openfunctions-v2):

```python
  if len(functions) == 0:
      return f"USER: <<question>> {user_query}\nASSISTANT: "
  functions_string = json.dumps(functions)
  return f"USER: <<question>> {user_query} <<function>> {functions_string}\nASSISTANT: "
```

PR #242, opened 2024-03-06
Local Inference script for openfunctions v2

ShishirPatil:
* Local inference for open-functions-v2.
* Fix: Error in how javascript output was parsed.


tianjunz (approved):
LGTM

PR #261, opened 2024-03-15
API Zoo Index Github Actions Fix

ramanv0:
This PR updates the github actions responsible for updating and validating API Zoo Index entries to handle very large files. The previously used curl requests were exceeding the max request size limit for large files, so the actions were changed to use a temporary file and the curl `--data-binary` option.


ShishirPatil (approved):
LGTM. Thanks for the quick fix @ramanv0

PR #268, opened 2024-03-17
Implement caching for API Zoo Index data

ramanv0:
This PR implements a "Stale While Revalidate" caching mechanism using the Service Worker and Cache APIs to improve API Zoo Index data fetching performance.

As can be seen in [this short demo](https://www.loom.com/share/637db9fb26ef461a86ee64a1ad5120ae?sid=e7183b42-e8e3-42a1-8c8d-0a9eb4d19740), the initial fetching of the data takes a few seconds, but the data for subsequent visits to the API Zoo Index webpage is cached, resulting in short load times. The caching strategy used returns the cached data immediately (if available) for a fast response, while an asynchronous fetch request runs in the background to update the cache.

How to test:
- clone the PR Changes
- navigate to the root directory of the project where the index.html file is located and run `python3 -m http.server`
- go to http://localhost:8000/ – this will serve the site from your local machine


ShishirPatil (approved):
LGTM.

PR #281, opened 2024-03-23
removed testing code

kaiwen129:
- changed `for chunk in chunks[:3]:` in `get_chunks` to `for chunk in chunks:` (originally kept to first 3 chunks for testing purposes)
- added instruction to the model to include only the questions in its response for `generate_instructions_gen`
- updated an inline comment to reflect the correct variable name
- removed a debug print statement


ShishirPatil (approved):
LGTM

PR #285, opened 2024-03-25
Cleaner Main Page

doranchang:
Big Changes:
• Abstracted majority of the CSS away from HTML index file
• Navigation bar works for all devices
• Citations text correctly lined up for all devices
------
Smaller Changes:
• Fixed edge and bottom margins for all devices
• Changed texts that would indent on certain devices
• Correctly centered videos and descriptions


assets/css/common-styles.css
```diff
@@ -17,9 +17,11 @@ body {
 .text-center {
     text-align: center;
     align-items: center;
+    margin-bottom: 2px !important;
```
HuanzhiMao, line 20: Please don't use `!important`. This will impact other pages as well, unless you change all other pages accordingly. Otherwise, please define another class that is used specifically for index page (and then you can do something like `class="text-center your-index-page-class"`).

index.html
```diff
@@ -34,8 +34,7 @@
         <a href="apizoo/">API Zoo Index</a>
     </div>
     <div class="highlight-clean" style="padding-bottom: 10px;">
-            <h1 class="text-center"> 🦍 Gorilla: Large Language Model Connected with Massive APIs</h1>
-        <div><p></p></div>
+        <h1 class="text-center" style="margin-left: 10px; margin-right: 10px;"> 🦍 Gorilla: Large Language Model Connected with Massive APIs</h1>
```
HuanzhiMao, line 37: This style is redundant when you already have the `!important` thing. Please try to only include the code that's needed.

index.html
```diff
@@ -84,19 +83,19 @@ <h3 class="text-center">Gorilla Used at
                 <div class="col-md-12 text-center"><img src="assets/img/gorilla_method.png" 
                     style="width: 80%;margin-bottom: 8px;" alt="Gorilla LLM logo">
                 </div>
-                <em>Gorilla is a LLM that can provide appropriate API calls. 
+                <p class="col-md-12" style="color: black;margin-bottom: 2px; margin-left: 10px; margin-right: 10px;">Gorilla is a LLM that can provide appropriate API calls.
```
HuanzhiMao, line 86: same here, why not just define a css rule for the `col-md-12` class in `main_page.css`? It has been repeated multiple times.

HuanzhiMao (changes requested):

PR #298, opened 2024-03-30
Add TreeMap to Leaderboard

HuanzhiMao:
This PR adds an interactive treemap to the website. Users can hover over and click on each block to see the detailed breakdown of error types for different models.


ShishirPatil (approved):
LGTM

PR #314, opened 2024-04-05
Support OSS Evaluation for Leaderboard 

### LINKED ISSUE #313: Open Source Model Eval
Hi, thanks for the amazing work!

Do we currently have support for evaluating OSS models? I've run `openfunctions_evaluation.py` however `eval_runner.py` does not parse `result.json`.

Thank you!

HuanzhiMao:
This PR adds support for OSS model evaluation for the Berkeley Function Calling Leaderboard. It will not change the score on the leaderboard. 

Close #313


berkeley-function-call-leaderboard/eval_checker/eval_runner_helper.py
```diff
@@ -697,6 +715,22 @@ def update_leaderboard_table_with_score_file(leaderboard_table, score_path):
                 }
 
 
+def oss_file_formatter(input_file_path, output_dir):
```
ShishirPatil, line 718: Can we do a simple assert on the # of rows in the file? If there is a mismatch, we should flag it.

ShishirPatil (changes requested):

berkeley-function-call-leaderboard/eval_checker/eval_runner_helper.py
HuanzhiMao: yea sounds good. Let me add that

ShishirPatil (approved):
LGTM.

PR #315, opened 2024-04-05
[Draft] Add support for Cohere command-r-plus

Fanjia-Yan:
This PR adds support for Cohere command-r-plus inference and output checking for Berkeley Function Calling Leaderboard. Both prompting and function calling modes are supported.

To inference fill up the Cohere API key and run function calling on command-r-plus with:
`python openfunctions_evaluation.py --model command-r-plus-FC`
You can also run our function calling via prompting to command-r-plus with:
`python openfunctions_evaluation.py --model comand-r-plus`

### DIFF
--- a/goex/README.md
+++ b/goex/README.md
@@ -15,27 +15,27 @@ GoEx provides a new way to interact with your favorite APIs! Powered by LLMs, Go
 
 ## Install
 
-1.) Navigate inside the goex folder and set up a clean environment with **Conda** or **Python venv**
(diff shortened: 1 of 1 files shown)


HuanzhiMao:
I think you forgot to update the cost (and maybe latency) info for the model.

harry-cohere:
Thanks a lot for the draft PR! I've submitted https://github.com/ShishirPatil/gorilla/pull/350 for consideration

PR #317, opened 2024-04-05
Leaderboard treemap data update

HuanzhiMao:
This PR updates treemap to remove Functionary-Small-v2.2 because we have included its newer v2.4 version. All other models are unaffected and data stay the same.
This PR **does not** change the leaderboard value.


CharlieJCJ (approved):
LGTM

PR #318, opened 2024-04-05
Support OSS Evaluation for Leaderboard

HuanzhiMao:
This PR adds support for OSS model evaluation for the Berkeley Function Calling Leaderboard. It will not change the score on the leaderboard. It is reopened from #314  because the GitHub server is down on April 4th and caused the PR to be unable to merge.

Close https://github.com/ShishirPatil/gorilla/issues/313


ShishirPatil (approved):
Clone of #314 since Github was down and commits since last night are still being processed in the PR.

PR #330, opened 2024-04-09
BFCL April 8th Release

HuanzhiMao:
This PR is for the leaderboard April 8th release:

1. Fixed an oversight that was introduced in #299. For function-calling (FC) models that cannot take `float` type in input, when the parameter type is a `float`, the evaluation procedure will convert that type to `number` in the model input and mention in the parameter description that `This is a float type value.`. An additional field `format: float` will also be included in the model input to make it clear about the type. 
2. Update the model handler for Claude, Mistral, and OSS to better parse the model output. This is to patch the handler we released in #299, as it sometimes fails to parse even though the model output is valid. This affects only the prompting models; the FC models are unaffected.

This PR **DOES** change the leaderboard score. We will update the leaderboard website shortly, in a different PR.

---------

Co-authored-by: Charlie Cheng-Jie Ji <charliechengjieji@berkeley.edu>
Co-authored-by: Fanjia Yan <fanjiayan@berkeley.edu>


Fanjia-Yan:
LGTM

CharlieJCJ (approved):
LGTM

ShishirPatil (approved):

PR #336, opened 2024-04-11
Initial goex commit

ShishirPatil:
We introduce GoEx, a runtime for LLMs with an intuitive `undo` and `blast-radius confinement` abstractions, enabling the safer deployment of LLM agents in practice. 

Paper: https://arxiv.org/abs/2404.06921


tianjunz (approved):
LGTM

PR #341, opened 2024-04-11
Leaderboard Update, in sync with BFCL April 9th Release

HuanzhiMao:
This PR updates the leaderboard data, as mentioned in #338. As a result, some values/scores are changed. 
Note that the model `glaiveai/glaive-function-calling-v1` is excluded in this leaderboard update PR due to the model's tokenizer problem. We cannot generate that model's output data. So to avoid confusion, we excluded that model in this update.


CharlieJCJ (approved):
LGTM

PR #343, opened 2024-04-11
Update requirements.txt

viniciuslazzari:
Running gorilla_falcon_cli.py withou einops package returns the following error:

ImportError: This modeling file requires the following packages that were not found in your environment: einops. Run `pip install einops`


ShishirPatil (approved):
Thanks for this @viniciuslazzari 🎉

PR #344, opened 2024-04-11
GoEx Gitignore User Credentials

royh02:
Make sure that user creds aren't version tracked.  Ex: `credentials/slack` needs to be ignored as well


Noppapon:
But we do want to track other files inside that folder, for example `supported.txt`. We should not use `*` and  only selectively untrack a subset of the files?

royh02:
Right. Added exlusion patterns and tested locally that changes in `credentials/credentials_utils.py` and `credentials/supported.txt` still show up, but everything else is still ignored.

Noppapon (approved):
Looks good to me! It might be worth moving credentials_utils.py and supported.txt, or the user's credentials, so they are in different folders. However, I think it's okay to merge this for now.

ShishirPatil (approved):
LGTM

PR #350, opened 2024-04-12
Add FC + Prompt for Cohere command-r-plus

harry-cohere:
Hello from Cohere 👋 

Thanks a lot for creating and maintaining this evaluation - it's been a lot of fun to work with, and the recent improvements are great to see.

This PR adds an initial implementation of Cohere's native tool use API (FC) and non-native prompt. It builds upon the work of  @Fanjia-Yan and @HuanzhiMao in https://github.com/ShishirPatil/gorilla/pull/315

I look forward to hearing what you think and seeing us on the leaderboard!


berkeley-function-call-leaderboard/model_handler/utils.py
```diff
@@ -10,6 +10,7 @@ def _cast_to_openai_type(properties, mapping, test_category):
     for key, value in properties.items():
         if "type" not in value:
             properties[key]["type"] = "string"
+            var_type = "string"
```
harry-cohere, line 13: These fixes are likely to change the leaderboard for LLMs that make use of the `mapping`, as it fixes a bug.

The bug is that we were trying to compare the type _after_ it was converted, so the compare always failed.

HuanzhiMao (approved):

Fanjia-Yan (approved):
Thanks Harry for pushing this! We will get the result up on the leaderboard for Command R Plus following this PR.

ShishirPatil (approved):
Thanks for this @harry-cohere 🎉

PR #353, opened 2024-04-14
Update raft.py with default `p` to match paper

### LINKED ISSUE #325: Hyperparameter of the number of distractor documents and the ratio of golden documents in training RAFT
Hi, I just wonder what the number of distractor documents and the ratio of golden documents in training RAFT are in producing the results in the paper. I saw the default ratio in the script the ratio is 1.0, which means no distractor documents are used. I wonder whether the default number in the scripts represents the hyperparameter used to produce the results in the paper?

ShishirPatil:
Change `p` which dictates the fraction of dataset with golden documents in them (vs) no golden documents. 

So, p = 0.8 means, for 80% of the train data set, `A* = Q + D* + D1 .. Dn` and for 20% of the train data set `A* = Q + D1 .. Dn` where `D*` are/is the golden document with the answer `A*`.

Close #325

### DIFF
--- a/raft/raft.py
+++ b/raft/raft.py
@@ -185,7 +185,7 @@ def add_chunk_to_dataset(
     doctype: DocType = "api", 
     x: int = 5, 
     num_distract: int = 3, 
-    p: float = 1.0
+    p: float = 0.8
 ) -> None:
     """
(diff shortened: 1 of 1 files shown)


tianjunz (approved):
LGTM

PR #369, opened 2024-04-18
Log error messages from API validation

eitanturok:
When you do API validation, we don't know which APIs fail. So I print out the APIs that fail so you know what is not working.


ShishirPatil (approved):
Thank you @eitanturok 💯

PR #370, opened 2024-04-18
Update .gitignore

eitanturok:
There are many files that are automatically generated when running BFCL that we should ignore. I update .gitignore to ignore:
1. the evaluation dataset that we download
2. the tree-sitter files in`berkeley-function-call-leaderboard/eval_checker` (right now we only ignore the symbolic link to the tree-sitter files in the `berkeley-function-call-leaderboard` directory)
3. The inference results in `berkeley-function-call-leaderboard/result/`
4. The evaluation results in `berkeley-function-call-leaderboard/score/`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,12 +10,45 @@ dist
 .editorconfig
 .DS_Store
 **/*.pyc
-./berkeley-function-call-leaderboard/function_credential_config.json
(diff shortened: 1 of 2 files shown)


HuanzhiMao (approved):
LGTM

CharlieJCJ (approved):
Look great! Thanks for your contribution @eitanturok

ShishirPatil (approved):

PR #373, opened 2024-04-19
Add missing argument to `OSSHandler`'s `_format_prompt` function

eitanturok:
# Overview

In this PR, I added a missing argument to `OSSHandler`'s `_format_prompt` function.

Also my code formatter sorted the imports alphabetically which accounts for all the other changes, including those in `handler_map.py`.

# What Happened?

In every model handler, we define `_format_prompt` as taking in three arguments (`prompt`, `function`, `test_category`). This is the standard. But in `OSSHandler`, `_format_prompt` is defined as only taking in `prompt` and `function` and forgetting to take `test_category` as an input.

This caused an error. So I simply added `test_category` to the definition of `_format_prompt`.


HuanzhiMao (approved):
Looks great! Thanks for contributing @eitanturok 
btw, which code formatter are you using? The import sorting feature seems handy :)

eitanturok:
It is a code formatter extension in VSCode: https://marketplace.visualstudio.com/items?itemName=ms-python.isort.

HuanzhiMao (approved):

Fanjia-Yan (approved):
LGTM, will not influence any of model generation output.

ShishirPatil (approved):

PR #381, opened 2024-04-23
Azure OpenAI support in raft.py

### LINKED ISSUE #382: Add Azure OpenAI support to RAFT's raft.py and eval.py
Currently, while it is possible to use Azure to run the fine-tuning, it is not currently possible to generate the fine-tuning dataset using Azure OpenAI.

Adding support for running the `raft.py` and `eval.py` scripts using Azure OpenAI deployments would be very useful for Azure users.

cedricvidal:
- Added `--embedding-model` CLI arg defaults to text-embedding-ada-002. Used to generate the chunks embeddings
- Added `--completion-model` CLI arg defaults to gpt-4. Used to generate the Q/A dataset
- New client_utils.py with two helpers: `build_openai_client` and `build_langchain_embeddings`.
- `build_openai_client` builds an `OpenAI` or `AzureOpenAI` client based on detecting `AZURE_*` specific env vars or not
- `build_langchain_embeddings` builds a `AzureOpenAIEmbeddings` or `OpenAIEmbeddings` based on detecting `AZURE_*` specific env vars or not
- Loads environment variables from .env file using python-dotenv package
- Added dependency on python-dotenv
- Added doc to README

Fixes #382


ShishirPatil (approved):
LGTM. Thanks @cedricvidal

PR #389, opened 2024-04-27
Gorilla Website Styling Update

HuanzhiMao:
This PR updates the appearance of the Gorilla website landing page. 

1. Add button links to the papers
2. Change the wording for the scrolling bar. 
3. Change nav bar ordering for the leaderboard to be consistant.

It **DOES NOT** change the leaderboard value


leaderboard.html
```diff
@@ -151,7 +151,8 @@ <h2>Leaderboard</h2>
             <p>
                 <b>Cost</b> is calculated as an estimate of the cost per 1000 function calls, in USD.
                 <b>Latency</b> is measured in seconds.
-                For <b>Open-Source Models</b>, the cost and latency are all calculated when serving with <a href="https://github.com/vllm-project/vllm">vLLM</a> using 8 V100 GPUs.
+                For <b>Open-Source Models</b>, the cost and latency are all calculated when serving with <a
```
CharlieJCJ, line 154: Also include Azure GPU pricing here for more clear documentation of prices.

leaderboard.html
```diff
@@ -51,10 +51,10 @@
         z-index: 100;
         font-size: 18px;
       ">
-        <a href="#leaderboard">Leaderboard</a>
-        <a href="#api-explorer">Try it Out!</a>
+        <a href="index.html">Home</a>
         <a href="blogs/8_berkeley_function_calling_leaderboard.html">Blog</a>
-        <a href="index.html">Gorilla</a>
+        <a href="#api-explorer">Try it Out!</a>
+        <a href="#leaderboard">Leaderboard</a>
```
CharlieJCJ, line 57: Can the last botton be api zoo index, rather than leaderboard, since we are currently in leaderboard (to be consistent with home page).
image

HuanzhiMao: We discussed this before. This is the leaderboard page, we want leaderboard here. It doesn't make sense to redirect to API zoo index.

HuanzhiMao: Same idea as why the blog in leaderboard redirect to blog 8, instead of the blog landing page.

leaderboard.html
HuanzhiMao: We have made it clear that we use Azure pay-as-you-go-price in the formula. You don't need to bother the audience with this level of detail; that's for the code. 
From the tweet we send:

> of-course costs vary with the contracts and vintage of GPUs, that is besides the point. With cost and latency, our goal is to identify what the magnitude of costs and latency looks like,  and understand the relative ordering when everything else is kept constant!

leaderboard.html
CharlieJCJ: Sure, can you also modify the navbar `Gorilla` to `Home` for https://huanzhimao.com/gorilla/blogs/8_berkeley_function_calling_leaderboard.html for consistency?

leaderboard.html
CharlieJCJ: There are people visiting the website without the context, we may link it at least to blog, or else this information is hard to find and confusing to audience. I had requests from twitter users.

leaderboard.html
HuanzhiMao: Sure. This I agree. We should let the blogs have consistent nav bar.

leaderboard.html
HuanzhiMao: We can add a line in the leaderboard say, the cost for oss is an estimate (rather than a precise number) to help identify what the magnitude of costs looks like.

CharlieJCJ: Sound good, go for it. Also link [blog 8](https://gorilla.cs.berkeley.edu/blogs/8_berkeley_function_calling_leaderboard.html#cost) for them to read for more details on cost latency calculations.

HuanzhiMao: But regarding the blog, telling the audience that the azure price is 20.3 vs 20.4 or 20.5 is like no difference. It's just some distracting information that doesn't help them understand the leaderboard values.

CharlieJCJ: No, I mean link a pointer for Azure ready-to-price, for which we get the numbers for estimate.

HuanzhiMao: Like think of it this way,
I told you, Azure price is 20.4, and you saw a model has cost 4.3. How does that help you determine the cost efficiency of that model if I told you Azure price is 21.4 instead? 
The absolute value doesn't matter. The relative ordering matters.

HuanzhiMao: Azure pay-as-you-go-price is just a unit for calculation. It could be AWS pay-as-you-go-price or GCP Azure pay-as-you-go-price, and it won't make any difference.

CharlieJCJ (approved):
LGTM

PR #403, opened 2024-05-01
Fix function names

eitanturok:
# The Problem

Many of the functions listed in `gorilla/berkeley-function-call-leaderboard/data/possible_answer` have extra symbols appended to the end of the function names, usually "_1", "_2", " 1", or " 2". 

For instance, in [gorilla_openfunctions_v1_test_parallel_multiple_function.json](https://github.com/ShishirPatil/gorilla/blob/4e8dac12515d847694bbfd4d1c15020ccf5bf263/berkeley-function-call-leaderboard/data/possible_answer/gorilla_openfunctions_v1_test_parallel_multiple_function.json#L3C1-L4C1) we see the function call
```python
{"circle.calculate_area": {"radius": [5]}, "circle.calculate_circumference_1": {"diameter": [10]}}
```
Why does `calculate_circumference_1` have a `_1` at the end of it? The user only has a function called `circle.calculate_circumference`, not a function called `circle.calculate_circumference_1`.

Here is another example:
```python
{"route_planner.calculate_route_1": {"start": ["New York"], "destination": ["Boston"], "method": ["fastest"]}, "chess_club_details.find_1": {"name": ["Knight Gambit"], "city": ["Boston"],"event":["null",""]}, "route_planner.calculate_route_2": {"start": ["Boston"], "destination": ["Philadelphia"], "method": ["fastest"]}, "chess_club_details.find_2": {"name": ["Rook Corner"], "city": ["Philadelphia"]}, "route_planner.calculate_route_3": {"start": ["Philadelphia"], "destination": ["New York"], "method": ["shortest"]}}
```
We call `chess_club_details.find` twice, once as `chess_club_details.find_1` and once as `chess_club_details.find_2`. However, the user only has access to the function `chess_club_details.find`, without the  `_1` or `_2`.

# My solution

Using the regex pattern 
```python
name = re.sub(r'(?:_|\s)\d+$', '', name)
```
I found 74 functions whose names had these extra characters just in the `gorilla/berkeley-function-call-leaderboard/data/possible_answer/gorilla_openfunctions_v1_test_parallel_multiple_function.json` file. 

In this PR, I removed these extra characters so that all the function names don't have a trailing number.

# To do

I'm not sure if this bug "matters" and just wanted to get some thoughts on this.

1. On one hand, I believe that fixing the function names should allow models to score higher on the BFCL. On the other hand, perhaps by using the AST parser, changing the function name does not actually effect the eval scores. I would love some guidance on this as I'm not sure if this fix will actually change the evals.
2. There are probably similar mistakes in the functions in `gorilla/berkeley-function-call-leaderboard/data/possible_answer/gorilla_openfunctions_v1_test_parallel_function.json`. These function names should be changed too.
3. One potential issue that may come up is that we store/read these function calls as json schema which is often compatible with dictionaries. In a dictionary, every key is unique. So if we use a function name as a dictionary key, then we will only be able to call that function once. If we'd call that function twice, then that dictionary would be using the same key twice and this would just overwrite the first entry.

Cheers!


HuanzhiMao (changes requested):
Hi @eitanturok,

Thanks for your contribution!
Two things I want to point out, 

1. We observe that some function calls seem to be accidently deleted in the possible answer, which would impact the evaluation pipeline's correctness. For example [here](https://github.com/ShishirPatil/gorilla/pull/403/files#diff-a77c22828a85cf7a75d42e9e94501a9c0d049afc7a2347d1fd775699f9dc8666L8-R8).
2. The third point you mentioned in the `TODO` section is indeed why we use suffix like `-1`, `-2` when the same function is invoked multiple times. These suffixes are necessary to ensure that the keys in a dictionary are unique when loading the possible answer JSON file. Without these, the two function calls would have the same key, and thus only show up as one function call after loading, which will be bad. 

I want to assure you that the suffix doesn't affect the correctness of our evaluation pipeline; we use the function name from the function doc for reference, not from the key in the possible answer. The code section are [here](https://github.com/ShishirPatil/gorilla/blob/main/berkeley-function-call-leaderboard/eval_checker/checker.py#L317-L339) for your reference. The line `possible_answer = list(possible_answer.values())[0]` essentially ignores the `key`, and only look at the `value` field of the possible answer for model result checking (and thus any suffix in the `key` won't matter). 

Let me know if you have more questions!

eitanturok:
Got it, that makes a lot of sense.

I'm happy to hear that the suffix does not actually the correctness of the pipeline. And thanks for explaining where in the code this takes place.

Might be helpful to emphasize that the ground truth function name does not really matter somewhere in the docs.

Closing this b/c my issue was addressed. Thanks for explaining this!

HuanzhiMao:
Yes, good point, we should explain that part somewhere (as other people might also have such concerns).

PR #406, opened 2024-05-03
handle parallel function calls from gemini

vandyxiaowei:
Handle parallel function calls for Gemini handler for the Berkeley Function Calling Leaderboard. 

This PR does NOT change values in BFCL.


HuanzhiMao (approved):
Thanks for contributing @vandyxiaowei! LGTM

Issue #363, opened 2024-04-16
Leaderboard evaluations issues


danieljannai21:
Hi,

First of all, I really appreciate your work!

I've been trying to run the leaderboard evaluation code, and noted several issues. I can probably fix some of them on my own in a PR, but wanted to hear your thought before that.

1. Specificity of argument description. In several functions, I noted that the description of the parameters is not explicit enough, which results in the models being evaluated sometimes outputting the "wrong" parameters. Attached are a few examples.

In this function, it isn't mentioned that `interest_rate` shouldn't be the number of percents (that is, 3.5 instead of 0.035 for example):
```python
def mortgage_calculator(loan_amount, interest_rate, loan_period):
    """
    Calculates the monthly mortgage payment.
    Args:
        loan_amount (integer): The amount of the loan.
        interest_rate (integer): The interest rate of the loan.
        loan_period (integer): The period of the loan.
    """
    monthly_interest_rate = interest_rate / 12
    number_of_payments = loan_period * 12
    monthly_payment = (
        loan_amount
        * monthly_interest_rate
        * (1 + monthly_interest_rate) ** number_of_payments
        / ((1 + monthly_interest_rate) ** number_of_payments - 1)
    )
    return monthly_payment
```

In this example, it isn't mentioned that the `stock_name` should be a ticker symbol and not the company's name:
```python
def get_company_name_by_stock_name(stock_name):
    """
    Finds the company name of a stock by its stock name.
    Args:
        stock_name (str): The stock name of the product.
    """
    url = "https://yahoo-finance15.p.rapidapi.com/api/v1/markets/search"

    querystring = {"search": stock_name}

    headers = {
        "X-RapidAPI-Key": api_key["RAPID-API-KEY"],
        "X-RapidAPI-Host": "yahoo-finance15.p.rapidapi.com",
    }

    response = requests.get(url, headers=headers, params=querystring)
    try:
        return response.json()["body"][0]["name"]
    except:
        return response.json()
```

2. Hard typing constraints.
For example, one of the questions in `gorilla_openfunctions_v1_test_executable_multiple_function.json` is `As a data analyst, you are working on a project that requires you to organize a set of numerical data. Can you sort these numbers 34, 2, 56, 7, 9, 12 in descending order?`, where the description of the relavant function is:
```json
{
  "name": "sort_array",
  "description": "Sorts an array of numbers.",
  "parameters": {
    "type": "dict",
    "properties": {
      "array": {
        "type": "array",
[...]

PR #407, opened 2024-05-03
better handle float value comparison

vandyxiaowei:
Refer discussion below.


HuanzhiMao (changes requested):
Hi @vandyxiaowei,

Thanks for the PR! 

However, we cannot merge it because the line `value = float(value)` will convert any value to a `float` when the parameter is expecting a float value, even if the original model-provided value is of type `int`. This compromises the purpose of type-checking in the AST evaluation process. 

The reason why we think strict typing is necessary can be found in [a previous issue reply](https://github.com/ShishirPatil/gorilla/issues/363#issuecomment-2076369788). The TLDR is that, without looking at the source code, it’s not 100% safe to say that the Python auto-conversion between float and int would always work as expected. Thus in the AST test category, we need to be precise with typing. There are certainly many scenarios where providing a float value for an integer parameter would work perfectly fine, and that’s exactly the purpose of the executable test category, where we are concerned only with the execution result (as opposed to how the result is obtained).  

PS, you might noticed that in line 397, we manually convert the value to `list` when the expected type is `tuple`, eg `value = list(value)` —- This is technically a workaround rather than best practice (and it does compromise the purpose of type-checking as well). The conversion is necessary there because any `tuple` in the possible answer would become a list after being processed through `json.dump()` and `json.load()`. I will have a more robust solution to address this issue :)

danieljannai21:
Hi @HuanzhiMao,

I just read again your response to the attached issue (which I opened) and I still disagree. I'm commenting here because the attached issue is already closed.

In your response you demonstrate why it's not okay to send a `float` to an argument that's expecting an `int` value, but the discussed scenario is the other way around -- sending an `int` value to a `float` argument, which I can't think of a reason of why it would be wrong. On top of that, in JSON schema, which is the standard way of representing a function in the API of OpenAI and others, the type used for representing floats is `number`, which also accepts integers (like 42 in [this example](https://json-schema.org/understanding-json-schema/reference/numeric#number) in the official documentation.

HuanzhiMao:
Hey @danieljannai21,

Sorry that issue was closed. I don't have admin access to re-open it, RIP:/ But you can always open a new follow-up issue :)

Regarding your first concern on why the Python auto-conversion from `int` to `float` would not always work:

The TLDR is that you cannot represent every `int` as `float` in Python. 
Python `int` objects have infinite/arbitrary precision (it is bounded only by memory). On the other hand, `float` objects do not have infinite precision; numbers are represented as an exponent and a significant, with the latter building up a number from 53 binary fractions. Thus, there are some `int` values that cannot be accurately represented by a `float` object. For example, consider a large int `809880023823263820`. Note how the value of the last two digits got changed from 20 to 72 after conversion. 
```
>>> x = 809880023823263820
>>> type(x)
<class 'int'>
>>> y = float(x)
>>> type(y)
<class 'float'>
>>> print(x)
809880023823263820
>>> print(y)
8.098800238232639e+17
>>> print(format(y, "f"))
809880023823263872.000000
```
Now, suppose I have the following function (you probably have seen such interest-rate-calculation-function in the BFCL dataset quite a lot haha):
```
def calculate_compound_interest(principal: float):
    """
    This function calculates the compound interest on a given principal over 10 periods.
    """
    for _ in range(10):
        principal += principal * 0.05
    return principal
```
Since we are talking about accuracy here, let's use a high-precision version of the above function. 
```
def calculate_high_precision_compound_interest(principal: float):
    """
    This function calculates the compound interest on a given principal over 10 periods, using Python's decimal module for high precision.
    """
    import decimal
    
    principal = decimal.Decimal(principal)
    rate = decimal.Decimal(0.05)
    for _ in range(10):
        principal += principal * rate
    return principal
```
The result is below. Sending an `int` value to a `float` argument results in an unexpected/wrong value. 
```
>>> large_int_principal = 809880023823263820
>>> large_float_principal = 809880023823263820.0
>>> print(calculate_high_precision_compound_interest(large_int_principal))
1319209219140100709.664993269
>>> print(calculate_high_precision_compound_interest(large_float_principal))
1319209219140100794.367513861
```


Now regarding your second concern that the OpenAPI standard allows float in `number` type:

1. Our evaluation dataset is not limited to the types defined in OpenAPI. We use the "BFCL Type", which for Python is `"boolean", "array", "string", "integer", "float", "tuple", "any", "dict"`. 
2. For models that accept OpenAPI input schema, the conversion from `float` to OpenAPI's `number` is similar to how we handle/convert the type for Java and JavaScript functions (a detailed explanation can be found in #424). 
3. Although the parameter type is the broad `number`, we specify the detailed type limitation in the parameter description. And thus the model should be capable of realizing such restrictions. The code that implements this is [here](https://github.com/ShishirPatil/gorilla/blob/main/berkeley-function-call-leaderboard/model_handler/utils.py#L17)


Let me know if you have more questions! Happy to discuss this further with you.

danieljannai21:
Hey @HuanzhiMao,

Thanks for taking the time to answer my concerns.

The problem in your example is that the number was inaccurately the moment it entered a `float` variable, and by the time it was converted to `decimal.Decimal` it was already the wrong number. The calculation that started with an `int` is actually closer to the correct result (which is `1319209219140100674.793173045685546875` according to [this calculator](https://www.mathsisfun.com/calculator-precision.html)). Regardless, all the examples in the evaluation suit are a lot smaller, so if this is indeed the reason you insist on matching the exact type, I think it would be best to add this type of edge case.

HuanzhiMao:
Hey @vandyxiaowei, @danieljannai21 and those who are following this issue,

After careful consideration and extensive internal discussions, we have decided to **allow** the use of `int` values for Python parameters expecting `float` values in BFCL. Here is our reasoning:

1. Python `int` is represented as C `long` (see [Python C API documentation](https://docs.python.org/3/c-api/long.html)). When converting from `int` to `float`, Python handles the conversion efficiently, even for large integers, by representing the `long` as $x \times 2^e$ where $0.5 \leq |x| < 1.0$  and  $e \geq 0 $. This ensures that rounding occurs to 53 significant bits, which means an `int` would have to be extraordinarily large to pose any problem (refer to the [CPython implementation](https://github.com/python/cpython/blob/main/Objects/longobject.c#L3470) and the special handling in the `_PyLong_Frexp` function).

2. Python’s internal handling of comparisons between `int` and `float` primarily concerns overflow and loss of precision (see [relevant code section](https://github.com/python/cpython/blob/main/Objects/floatobject.c#L391-L404)). According to the [code comment block](https://github.com/python/cpython/blob/main/Objects/floatobject.c#L391-L404), this behaviour does not introduce any additional issues beyond those inherent to floating-point arithmetic.

3. The `float` parameter values used in BFCL and typical real-life function-calling scenarios are relatively small and well within the limits where conversion issues like overflow or significant precision loss are relevant. Therefore, in practical terms, the potential problems are negligible.

Given these points, we believe that allowing `int` values for `float` parameters is a reasonable and safe decision. We appreciate all the discussion and insights from the community. We will update the score on the leaderboard very soon (in a different PR).

Thank you all for your contributions!
Best,
Huanzhi

HuanzhiMao (approved):

ShishirPatil (approved):
LGTM.

PR #415, opened 2024-05-08
RAFT Fix arbitrary code execution vulnerability in checkpoint feature

cedricvidal:
Replaced `eval()` by `int()` when loading the checkpoint file to remove the possibility of executing arbitrary code.


ShishirPatil (approved):
Thanks for catching the bug.

PR #419, opened 2024-05-09
eval fixes

devashishtyagi:


berkeley-function-call-leaderboard/model_handler/firework_ai_handler.py
```diff
@@ -7,7 +7,9 @@
 class FireworkAIHandler(OpenAIHandler):
     def __init__(self, model_name, temperature=0.7, top_p=1, max_tokens=1000) -> None:
         super().__init__(model_name, temperature, top_p, max_tokens)
-        self.model_name = "accounts/fireworks/models/firefunction-v1-FC"
+        #self.model_name = "accounts/fireworks/models/dt-fc-rc-v3-FC"
```
pgarbacki, line 10: Lets revert this for easier merging with upstream in the future.

berkeley-function-call-leaderboard/model_handler/gpt_handler.py
```diff
@@ -1,32 +1,37 @@
...
-    USER_PROMPT_FOR_CHAT_MODEL,
-    SYSTEM_PROMPT_FOR_CHAT_MODEL,
 )
 from openai import OpenAI
-import os, time, json
 
 
 class OpenAIHandler(BaseHandler):
-    def __init__(self, model_name, temperature=0.7, top_p=1, max_tokens=1000) -> None:
+    def __init__(self, model_name, temperature=0.0, top_p=1, max_tokens=1000) -> None:
```
pgarbacki, line 24: Move this to `FireworkAIHandler`

berkeley-function-call-leaderboard/model_handler/gpt_handler.py
```diff
@@ -1,32 +1,37 @@
...
         super().__init__(model_name, temperature, top_p, max_tokens)
         self.model_style = ModelStyle.OpenAI
         self.client = OpenAI(api_key=os.getenv("OPENAI_API_KEY"))
 
-    def inference(self, prompt,functions,test_category):
+    def inference(self, prompt, functions, test_category):
```
pgarbacki, line 29: Shall we overwrite it in FireworksAIHandler? If we publish our results, we will likely need to send a PR against upstream. I would prefer to keep the changes in the fireworks-specific code.

berkeley-function-call-leaderboard/model_handler/handler_map.py
```diff
@@ -47,6 +47,7 @@
     "mistral-small-2402-FC-Auto": MistralHandler,
     "mistral-tiny-2312": MistralHandler,
     "fire-function-v1-FC": FireworkAIHandler,
+    "dt-fc-rc-v3": FireworkAIHandler,
```
pgarbacki, line 50: Revert.

pgarbacki (changes requested):
I would suggest the following:
* separate out the changes that are related to syncing with the upstream (READMEs, etc). Lets have only the relevant changes here so we can more easily track them in the future,
* lets try to put all the changes in `firework_ai_handler.py`. In the future, we will put a PR with these changes against the upstream.

PR #420, opened 2024-05-09
RAFT Reverted `raft.py` param names to dash-case

### LINKED ISSUE #414: RAFT raft.py CLI arguments naming broke `--embedding-model` and `--completion-model` in PR #410
**Is the feature request related to a problem?**
PR #410 introduced a breaking change in the CLI arguments names:
- `--embedding-model` was renamed to `--embedding_model`
- `--completion-model` was renamed to `--completion_model`

It seems like this was done to align to the double dash prefixed snake-case naming convention used with the other `--chunk_size` argument.

Here are the changed lines:
https://github.com/ShishirPatil/gorilla/commit/624d371e085eefd23ae3f818107490ee8f59e8d9#diff-b81150e1021ea386316094b5858fd74e2cbc1b1f1acc1dc37ed63d1507c87727L38-R44

**Describe the solution you'd like**
Having a consistent naming convention is important but the double dash prefixed snake case convention is quite unusual for CLI arguments. They usually use the dash-case naming convention.

While this is not a big deal, I would vote to change all arguments to dash-case like so:
- `--embedding-model`
- `--completion-model`
- `--chunk-size`

cedricvidal:
Fixes #414

### DIFF
--- a/raft/raft.py
+++ b/raft/raft.py
@@ -40,8 +40,8 @@ def get_args() -> argparse.Namespace:
     parser.add_argument("--chunk_size", type=int, default=512, help="The size of each chunk in number of tokens")
     parser.add_argument("--doctype", type=str, default="pdf", help="The type of the document, must be one of the accepted doctypes", choices=["pdf", "txt", "json", "api"])
     parser.add_argument("--openai_key", type=str, default=None, help="Your OpenAI key used to make queries to GPT-3.5 or GPT-4")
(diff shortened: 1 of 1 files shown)


cedricvidal:
@kaiwen129 fyi

kaiwen129 (approved):

kaiwen129:
@cedricvidal Should we keep things consistent and change `chunk_size` and `openai_key` in the args list to hyphen case as well?

cedricvidal:
Yes of course, I did this quickly and missed it

Cedric

PR #427, opened 2024-05-15
Leaderboard Update, in sync with PR#406 (handle parallel function calls from gemini)

HuanzhiMao:
This PR updates the leaderboard to reflect the change in score due to the added support for parallel function calls for gemini series in #406.


CharlieJCJ (approved):
Checked numbers, LGTM

PR #439, opened 2024-05-22
[BFCL] Standardize Model Name Among handler_map and eval_runner_helper

HuanzhiMao:
This PR standardize the model name among `handler_map` and `eval_runner_helper`. In the past, the `handler_map` used the model name that contains `/` while the `eval_runner_helper` takes the model name that has the `/` substituted with `_`. This caused a lot of unnecessary confusion when trying to add new models, so this PR addresses it by letting the `eval_runner_helper` adapt to the `handler_map` format.
This PR **DOES NOT** affect the leaderboard score.


HuanzhiMao:
Edit, we will do it the other way around. We will switch to the version where the model name contains `/`, because the model name that people copy directly from huggingface will contain `/`, so that's more convenient.

Fanjia-Yan (approved):
LGTM

ShishirPatil (approved):

PR #449, opened 2024-05-29
Fix indentation in leaderboard README

polm-stability:
The code examples in the README use both triple backticks and indentation, which is not necessary and makes the code examples appear excessively indented. This is also annoying if you use the bash `HISTCONTROL=ignorespace` option, which causes commands prefaced with a space to not appear in history.

This PR removes the extra indentation.


HuanzhiMao (approved):
Thanks for contributing @polm-stability!

ShishirPatil (approved):
Thank you @polm-stability

PR #452, opened 2024-05-30
Fix breaking changes due to updated Anthropic SDK

eitanturok:
Anthropic just moved their tool use from beta to main so we have to change the import `from anthropic.types.beta.tools import ToolUseBlock` to `from anthropic.types import ToolUseBlock`. You cannot run the eval without this change as things break.

Also, my IDE automatically sorted the imported packages and removed some extra spaces -- this explains all the other changes.

### DIFF
--- a/berkeley-function-call-leaderboard/model_handler/claude_fc_handler.py
+++ b/berkeley-function-call-leaderboard/model_handler/claude_fc_handler.py
@@ -1,18 +1,20 @@
-from model_handler.handler import BaseHandler
+import json
+import os
+import time
+
 from anthropic import Anthropic
-from anthropic.types import TextBlock
-from anthropic.types.beta.tools import ToolUseBlock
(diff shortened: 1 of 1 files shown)


eitanturok:
@justinwangx, @HuanzhiMao

HuanzhiMao (approved):
Thanks for the PR @eitanturok!

@ShishirPatil LGTM, ready to merge.

eitanturok:
Also, this situation would be avoided if you pin your package versions. As a best practice, might be worthwhile doing this...

HuanzhiMao:
> Also, this situation would be avoided if you pin your package versions. As a best practice, might be worthwhile doing this...

Yes, good point. We will add that.

ShishirPatil (approved):
LGTM. Thanks @eitanturok

PR #462, opened 2024-06-08
[BFCL] PR#407 Evaluation Pipeline Robustness Patch

HuanzhiMao:
This PR makes the BFCL evaluation pipeline more robust to the changes introduced in #407: When the parameter is of type `float`, less-capable models sometimes output a `string` or other types that might not be suitable to be converted to a `float`, which will cause the evaluation pipeline to error in `float(value)`. To prevent this, the pipeline now verifies the `value` type before attempting conversion to `float`.
This PR **DOES NOT** change the leaderboard evaluation score.

### DIFF
--- a/berkeley-function-call-leaderboard/README.md
+++ b/berkeley-function-call-leaderboard/README.md
@@ -234,7 +234,7 @@ For inferencing `Databrick-DBRX-instruct`, you need to create a Databrick Azure 
 
 
 ## Changelog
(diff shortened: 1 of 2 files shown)


Fanjia-Yan (approved):
Confirmed that the Leaderboard has been updated in #459 for the PR code changes.

ShishirPatil (approved):

PR #437, opened 2024-05-20
Fixes For NexusHandler

VenkatKS:
Hello! 

I just had a chance to take a look at the handler code for one of the open source models for Raven, and noticed a few things that could be fixed. I have made some fixes here, including fixes for converting the typescript Function Calling DSL into Python Function Calling DSL. This PR is for fixing the prompting issues only.

I believe this logic can be used for other Prompt-based models as well. I notice you ask the model to generate Python outputs but give JSON inputs for prompt-based FC; this can be applied to standardize the prompting side for these other models as well.

I believe there are further fixes necessary on the AST parsing of the outputs that can further improve the performance of the model. Happy to help contribute more there!

I ran the model with these changes, and these are the new results I notice: 

| Category        | REPORTED | UPDATED |
|-----------------|----------|---------|
| Simple AST      | 68.91    | 80.5    |
| Multiple AST    | 76       | 86      |
| Parallel AST    | 30.5     | 42      |
| Parallel Multiple | 43.5    | 49      |
| Simple Exec     | 64.12    | 95      |
| Multiple Exec   | 82       | 92      |
| Parallel Exec   | 46       | 74      |
| Parallel Multiple I | 55    | 62.5    |
| Relevance       | 2.08     | 57.5    |
| **AVG**         | **52.01222222** | **70.94444444** |

Here are the results JSONs for each of the above with the new runs:
1. Simple AST: [simple_score.json](https://github.com/ShishirPatil/gorilla/files/15380677/simple_score.json)
2. Multiple AST: 
[multiple_function_score.json](https://github.com/ShishirPatil/gorilla/files/15380683/multiple_function_score.json)
3. Parallel AST: 
[parallel_function_score.json](https://github.com/ShishirPatil/gorilla/files/15380686/parallel_function_score.json)
4. Parallel Multiple AST: 
[parallel_multiple_function_score.json](https://github.com/ShishirPatil/gorilla/files/15380691/parallel_multiple_function_score.json)
5. Simple Exec: 
[executable_simple_score.json](https://github.com/ShishirPatil/gorilla/files/15380694/executable_simple_score.json)
6. Multiple Exec: 
[executable_multiple_function_score.json](https://github.com/ShishirPatil/gorilla/files/15380696/executable_multiple_function_score.json)
7. Parallel Exec: 
[executable_parallel_function_score.json](https://github.com/ShishirPatil/gorilla/files/15380699/executable_parallel_function_score.json)
8. Parallel Multiple Exec: 
[...]

PR #472, opened 2024-06-14
Leaderboard Update, in sync with PR#437 (Fixes For NexusHandler)

HuanzhiMao:
This PR updates the score on the leaderboard to reflect the updated `NexusHandler` from #437. This handler only affects `Nexusflow-Raven-v2`, so the score for all other models remain unchanged. 
We want to thank @VenkatKS for patching the `NexusHandler`.


CharlieJCJ (approved):
LGTM

PR #470, opened 2024-06-13
Add firefunction-v2 to the leaderboard

pgarbacki:
This PR adds firefunction-v2 to the leaderboard. It also updates the fireworks handler making sure that the parameters and prompt are set correctly.

PR #476, opened 2024-06-19
Leaderboard Update, adding new model firefunction-v2-FC

HuanzhiMao:
This PR adds the new model `firefunction-v2-FC` to the leaderboard, thanks to the support from @pgarbacki in #470.
This PR only adds an entry in the leaderboard; the score for all other models on the leaderboard will remain unchanged.


dzhulgakov:
@HuanzhiMao - do you need anything else to update the leaderboard on the website? #470 landed, it'd be awesome to get Firefunction V2 numbers on the leaderboard. Thanks!

HuanzhiMao:
Hey @dzhulgakov, we are finalizing the results. One of the executable APIs is not stable, and that's why the leaderboard update is delayed a bit. It would be out very very soon.

CharlieJCJ (approved):
LGTM

PR #487, opened 2024-06-26
Added Agent Marketplace to Gorilla Repository

NithikYekollu:
Added [agent-marketplace repository](https://github.com/gorilla-llm/agent-marketplace) in the agent-marketplace directory within the Gorilla repository.


ShishirPatil:
Hey @NithikYekollu thanks for the PR. I am wondering if we should break this down into 2 parts. One part here part of main, and the other website stuff in gh-pages branch? Although given most of the logic is in the backend server, and most of the code here is front-end, we could just choose main or gh-pages brach. Thoughts?

NithikYekollu:
Hey Shishir, appreciate the feedback. Due to the website having non static components, and being hosted through vercel, I thinkt he gh-pages option wouldn't work currently

PR #489, opened 2024-06-26
[BFCL] Add ability to evaluate Nemotron-4-340B-Instruct

Fanjia-Yan:
This PR will add ability to inference on Nemotron via Nvidia NIM. It's open source and hosted by Nvidia without specifying the price per inference call. Therefore, no cost per thousand function calls will be recorded. Model result will be updated in a separate PR.


HuanzhiMao (approved):
LGTM

PR #496, opened 2024-07-02
[BFCL] Sanity check is now optional

### LINKED ISSUE #486: [BFCL] Sanity check should be optional and by default off
For BFCL eval, the sanity check of weather all RESTful APIs are active should be optional flag that is `off` by default. It is currently `on` by default.

ShishirPatil:
1. Sanity check is now optional. Further if the checks fail, we catch that and then continue the evaluation. That way if an API is down, it doesn't hinder other evaluations especially AST evaluations.
2. Update README.

Close #486 

This PR does NOT update the Leaderboard values.


HuanzhiMao (approved):
Good to go!

Issue #501, opened 2024-07-06
BFCL setup instruction is very difficult to follow


HuanzhiMao:
We need to make it much more user-friendly!

shizhediao:
Is there a way to evaluate BFCL without using vllm? for example, can I use huggingface transformer's generate()?

PR #504, opened 2024-07-06
[BFCL] Add Test Dataset to Repository

HuanzhiMao:
Previously, the test dataset was stored on HuggingFace, requiring users to clone this repository, download the dataset separately from HuggingFace, and then run the evaluation pipeline. This has caused many inconveniences and confusion within the community. Users often struggled with the inconsistency of having the possible answers within the repository while the test dataset was missing.

Partially addresses #501.


.gitignore
```diff
@@ -28,24 +28,8 @@ berkeley-function-call-leaderboard/eval_checker/tree-sitter-javascript
 berkeley-function-call-leaderboard/tree-sitter-java
 berkeley-function-call-leaderboard/tree-sitter-javascript
 
-# Ignore Evaluation Dataset files that we download from https://huggingface.co/datasets/gorilla-llm/Berkeley-Function-Calling-Leaderboard/tree/main
+# Ignore aggregated eval data (used for OSS models)
```
ShishirPatil, line 31: Why so? Why not also include OSS evals?

ShishirPatil (approved):

.gitignore
HuanzhiMao: `eval_data_total.json` is generated by `eval_data_compilation.py`, which just compiles all the test files into one. There is no need to keep track of it, since it can (and should) be generated anew on each run (to accommodate any possible changes to the dataset.

Issue #502, opened 2024-07-06
Clarify Documentation About Running The Benchmark


hamelsmu:
I had trouble following [the instructions for running the benchmark](https://github.com/ShishirPatil/gorilla/tree/main/berkeley-function-call-leaderboard/README.md)

1.  Why do we have to apply the credentials one file at a time?  Don't all the files get concatenated by `eval_data_compilation.py`?

> ![image](https://github.com/ShishirPatil/gorilla/assets/1483922/78763280-a8db-46dd-bb3e-709fb1fc0176)

2.  The TEST_CATEGORY is not the same thing between `openfunctions_evaluation.py` and `eval_checker/eval_runner.py` ?  Why are the test categories different?

> ![image](https://github.com/ShishirPatil/gorilla/assets/1483922/bc6b7d15-d77c-4b97-b9f4-b3e10aa5e9a1)

3.  Should the `MODEL_NAME` be the same between `openfunctions_evaluation.py` and `eval_checker/eval_runner.py` ?  or is there a reason you might want them to be different?

It would be helpful to more fully describe what each step is doing and why if possible.  Also, another question is the `OMDB API` is really flaky in my experience, I intermittently get 401 errors for no reason, even on the same request (sometimes 200, sometimes 401).  But I couldn't figure out how to turn test cases that involved this API off, and it caused things to fail.  Perhaps by understanding what each step is meant to do in more detail it will help me run the benchmark!  

cc: @HuanzhiMao

HuanzhiMao:
Hi @hamelsmu,

Thank you for raising this issue. We have addressed all the concerns you mentioned. Sorry for the delay in the merging process.

- PR #508 and #512 address your first concern.
- PR #506 addresses your second concern.
- PR #439 addresses your third concern.
- Regarding the reliability of the API, we are monitoring it closely and will consider replacements if any continue to be unreliable. Currently, they are all functioning properly. Further, the API sanity check has been set to be off by default in PR #496.

PR #506, opened 2024-07-06
[BFCL] Standardize TEST_CATEGORY Among eval_runner.py and openfunctions_evaluation.py

HuanzhiMao:
There are inconsistencies between the `test_category` argument that's used by `eval_checker/eval_runner.py` and `openfunctions_evaluation.py`. 

This PR partially addresses #501 and #502.


CharlieJCJ (approved):
LGTM

Comment: improve error message, make them more information for user to know immediate action items when aggregating results for `data.csv`.

HuanzhiMao:
> LGTM
> 
> Comment: improve warning message in the future, make them more information for user to know immediate action items when aggregating results for `data.csv`.

Good point. Will address this in a different PR.

ShishirPatil:
Hey @HuanzhiMao and @CharlieJCJ I don't think this is a great idea for the following reason. So we start by showing how to install the dependencies, and then in the middle we go into a long extempore on the different flags and then we come back with a list of commands to run. This ain't helpful back and forth? Maybe we retain the previous structure where after each command we show the options so users have the context on the command. Thoughts?

HuanzhiMao:
> Hey @HuanzhiMao and @CharlieJCJ I don't think this is a great idea for the following reason. So we start by showing how to install the dependencies, and then in the middle we go into a long extempore on the different flags and then we come back with a list of commands to run. This ain't helpful back and forth? Maybe we retain the previous structure where after each command we show the options so users have the context on the command. Thoughts?

How about we move the flags section to the bottom of the README, just like how all the available model names are displayed at the end?

devanshamin:
I have solved this issue [here](https://github.com/ShishirPatil/gorilla/pull/510/commits/c7c51672a5516553e3af32714ee134985504ac52).

ShishirPatil:
> > Hey @HuanzhiMao and @CharlieJCJ I don't think this is a great idea for the following reason. So we start by showing how to install the dependencies, and then in the middle we go into a long extempore on the different flags and then we come back with a list of commands to run. This ain't helpful back and forth? Maybe we retain the previous structure where after each command we show the options so users have the context on the command. Thoughts?
> 
> How about we move the flags section to the bottom of the README, just like how all the available model names are displayed at the end?

I don't think this solves it right? We should present the flags when the user cares about the command, not super down. Folks read and execute code from READMEs linearly, and making them jump around isn't a good idea imo.

ShishirPatil (approved):

Issue #477, opened 2024-06-19
Question about AST evaluation for Java and JavaScript


lucenzhong:
https://github.com/ShishirPatil/gorilla/blob/2f39693df9868737e8074091d549b652d446d995/berkeley-function-call-leaderboard/model_handler/gpt_handler.py#L94C1-L109C30
Hello, May I ask why the gpt model with FC do not use `tree-sitter` to do the AST parsing for Java and JavaScript test categorie ?

HuanzhiMao:
Hi @lucenzhong,
GPT models in FC mode don't need to have their result parsed for Java and JavaScript categories, because the model output is in JSON format and all the parameters involved with the Java/JS test are of type `string`, which is JSON compatible. So here, `json.load` is enough to get the result ready to be fed into the evaluation pipeline. The actual parsing part (to turn parameter value from `string` into their 'real' type) for these parameters happens in the checker section([here](https://github.com/ShishirPatil/gorilla/blob/main/berkeley-function-call-leaderboard/eval_checker/checker.py#L364-L388)), which involves calling the [Java type converter](https://github.com/ShishirPatil/gorilla/blob/main/berkeley-function-call-leaderboard/eval_checker/java_type_converter.py) and the [JS type converter](https://github.com/ShishirPatil/gorilla/blob/main/berkeley-function-call-leaderboard/eval_checker/js_type_converter.py). These two converters both make use of the `tree-sitter`.
Let me know if you have more questions!

PR #516, opened 2024-07-08
[BFCL] Fix Double-Casting Issue in model_handler for Java and JS category. 

HuanzhiMao:
When inferencing for Java or JS category, all function parameters are already in `string` type (refer to [this part](https://github.com/ShishirPatil/gorilla/blob/main/berkeley-function-call-leaderboard/model_handler/utils.py#L334-L374)). So the model output must be `string` type as well to be correct. 

The `decode_ast` method in some of the `model_handler` contains the following code snippet:
```
if language == "Python":
    pass
else:
    # all values of the json are casted to string for java and javascript
    for key in params:
        params[key] = str(params[key])
```
The line `params[key] = str(params[key])` should not exist because it is giving the model a 'second chance' when its parameter type is wrong (by casting it to a `string` when it is not). This double type-casting will introduce false positives to the evaluation result. As I mentioned in #477, this PR addresses this issue. 

Model Family Affected:

- OpenAI (FC only)
- Claude (FC only)
- Mistral (FC only)
- Gemini
- Cohere (FC only)

This will affect the leaderboard score. We will update it soon, in a different PR.


Fanjia-Yan (approved):
LGMTM

ShishirPatil (approved):

PR #525, opened 2024-07-12
[BFCL] add ibm-granite-20b-functioncallling model

MayankAgarwal:
This PR adds the [granite-20b-functioncalling](https://huggingface.co/ibm-granite/granite-20b-functioncalling) model to the BFCL.

This PR only adds the model handler. The leaderboard should be updated in a separate PR.

### DIFF
--- a/berkeley-function-call-leaderboard/README.md
+++ b/berkeley-function-call-leaderboard/README.md
@@ -194,6 +194,7 @@ Below is *a table of models we support* to run our leaderboard evaluation agains
 |snowflake/arctic | Prompt|
 |nvidia/nemotron-4-340b-instruct| Prompt|
 |THUDM/glm-4-9b-chat 💻| Function Calling|
+|ibm-granite/granite-20b-functioncalling 💻| Function Calling|
 
 Here {MODEL} 💻 means the model needs to be hosted locally and called by vllm, {MODEL} means the models that are called API calls. For models with a trailing `-FC`, it means that the model supports function-calling feature. You can check out the table summarizing feature supports among different models [here](https://gorilla.cs.berkeley.edu/blogs/8_berkeley_function_calling_leaderboard.html#prompt).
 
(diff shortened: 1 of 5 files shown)


HuanzhiMao:
Thank you for the PR @MayankAgarwal ! We will review and test it, and update the leaderboard accordingly.

HuanzhiMao (approved):
Hey @MayankAgarwal, 

I noticed that you intentionally removed the language-specific hint we included at the end of the model prompt [here](https://github.com/ShishirPatil/gorilla/pull/525/files#diff-64c3e18612454d570980d67c9ddfdeb2195acd2538e68d7630205fbda2388ebdR28-R33). I'm wondering what's the reasoning behind this change? 

In my opinion, this removal could negatively impact the model's performance, especially in the Java and JavaScript test categories. However, we are fine with the removal as it won't unfairly benefit the model. I just want to bring this to your attention.

The rest of the PR looks great!

MayankAgarwal:
Hi @HuanzhiMao,

Thank you for reviewing and approving the PR. 

We chose to remove the language-specific prompt augmentation because of the following reasons:
1. Since we unified the input and output format to JSON for this model, we were not sure of including any language-specific hint in the prompt. I believe that unless the model outputs code in that specific language (say Java or JavaScript in the case of BFCL), there should be a separate process to convert the function-calling output from a model to that specific language, rather than the model itself generating language-compliant code in JSON/XML data format.
2. We did not add the language-specific hint in the prompt while experimenting with other tasks and datasets. Thus, to maintain consistency we did not include it in this PR, and
4. While building the model handler, we experimented with both adding and removing the language-specific hint and we did not see a significant difference in the performance.

Owing to these reasons, we decided to remove the language-specific hint. I hope these sound reasonable to you.

If there are no changes from your side, please merge the PR. I would love to see the model on the leaderboard.

Thanks again for reviewing the PR!

ShishirPatil:
Thank you for the PR @MayankAgarwal  and delighted to welcome IBM models into the BFCL!

PR #531, opened 2024-07-18
fix line return

fantasist:
`len(data)` is the size of json object parsed from a line, which does not represent the count of lines in the source file. This incorrectly squashes most contents of the input file to the second line of the output file.


HuanzhiMao (approved):
Good catch. Thanks for the PR @fantasist!

fantasist:
Hi @HuanzhiMao , I don't have permission to merge. Feel free to merge it for me when you get a chance.

ShishirPatil:
Thanks for the catch @fantasist and welcome!

Issue #424, opened 2024-05-11
Java/Javascript split questions


vandyxiaowei:
1. can we show the detailed scores for Java/Javascript splits on the leaderboard (instead of the averaged Simple function)? One observation I have from running the benchmark myself is that Prompt model seems to have more advantages over FC endpoints. 
2. for Java types (Array/ArrayList/Hashmap), from the parsing code here: https://github.com/ShishirPatil/gorilla/blob/ae5f0a2daeed5a7c53ff8efda143dee55ecb17e9/berkeley-function-call-leaderboard/eval_checker/java_type_converter.py#L37 are we expecting the parameter values are in the form of "new ArrayList", "new HashMap", etc. I think usually users' expectations on FC endpoint outputs (not Prompt) are JSON object. 

Thanks!

HuanzhiMao:
Hi @vandyxiaowei,

Regarding your first question: 
Good point. We will display the detailed score breakdown of the AST Simple category in the next website release (likely next week on May 16/17, as I'm super busy till the 15th, rip).

Regarding your second question: 
Yes, we are expecting values in the format of `new ArrayList`; it needs to be valid Java syntax, and a JSON list `[]` would not be correct. 
For Java (and JavaScript as well), before querying the model, we do some [pre-processing on the prompt and function document](https://github.com/ShishirPatil/gorilla/blob/main/berkeley-function-call-leaderboard/model_handler/gemini_handler.py#L88-L90). Specifically, at the end of the prompt, we will explicitly state that `the provided function is in Java 8/JavaScript/Python syntax`. And for parameter types that are not native to JSON, we will change their type to `String` (since `String` is JSON compatible) and add in the parameter description that `"This is Java/Javascript" + {original_type} + " in string representation."`
So in the example you provided above, when expecting type `ArrayList`, model will get the instruction that this is a `String` type parameter with the parameter description containing `"This is Java ArrayList in string representation."`, and thus the model should output the value in `String` format (eg, `"new ArrayList<>(Arrays.asList(10, 20, 30))"`), which is JSON compatible. 
The relevant code that handles the pre-processing is [here](https://github.com/ShishirPatil/gorilla/blob/main/berkeley-function-call-leaderboard/model_handler/utils.py#L334-L374).

Let me know if you have more questions!

GeniusYx:
Hi, where can I see the score of each individual item of AST Simple?

Thanks!

GeniusYx:
[...]

PR #538, opened 2024-07-21
[BFCL] Fix language_specific_pre_processing for Java and JavaScript Test Category

HuanzhiMao:
In our BFCL official communication channels, including the evaluation manual blog, GitHub issue replies (such as #424), and our Discord channel, we have previously stated the following:

> For Java and JavaScript test category, before querying the model, we do some pre-processing on the prompt and function document. Specifically, at the end of the prompt, we will explicitly state that `the provided function is in Java 8/JavaScript/Python syntax`. And for parameter types that are not native to JSON, we will change their type to `String` (since `String` is JSON compatible) and add in the parameter description that `f" This is Java/JavaScript {value['type']} in string representation."`
> As an example, when expecting type `ArrayList`, model will get the instruction that this is a `String` type parameter with the parameter description containing `"This is Java ArrayList in string representation."`, and thus the model should output the value in `String` format (eg, `"new ArrayList<>(Arrays.asList(10, 20, 30))"`), which is JSON compatible. 

However, the code for `language_specific_pre_processing` did not implement this correctly. Due to an indentation issue, the parameter description was only modified when the parameter type was `any`, and the part where the parameter type is cast to `String` was never implemented. 

This issue was unnoticed until PR #516 was merged because of the double-casting problem. 

It *significantly impacts* the evaluation score for the Java and JavaScript categories. We will update the leaderboard very soon.

This PR:
- Addresses the above issue and ensures that the evaluation logic aligns with the previously described behaviour
- Updates two entries in the JavaScript dataset, due to their parameters missing a `description` field
  - Index: `14, 45`

We sincerely apologize for the oversight.


Fanjia-Yan (approved):
lgtm

ShishirPatil (approved):

PR #512, opened 2024-07-08
[BFCL] Support Category-Specific Generation for OSS Model, Remove eval_data_compilation Step

HuanzhiMao:
Currently, for OSS models, users must run `eval_data_compilation.py` to merge all datasets into a single `data_total.json` file before running inference. This requires inferring on all datasets at once, without the option to run inference on individual datasets or subsets. This PR addresses this limitation by allowing users to perform inference on specific datasets directly, removing the need for the `eval_data_compilation` step.
Note: hosted models don't have this limitation. 

Partially addresses #501 and #502.

PR #540, opened 2024-07-22
[BFCL] Support Multi-Model Multi-Category Generation; Add Index to Dataset; Handle vLLM Benign Error

HuanzhiMao:
In this PR:

1. **Support Multi-Model Multi-Category Generation**:
   - The `openfunctions_evaluation.py` can now take a list of model names and a list of test categories as command line input. 
   - Partially address #501.

2. **Handling vLLM's Error**:
   - A benign error would occur during the cleanup phase after completing a generation task, causing the pipeline to fail despite generating model results. This issue stems from vLLM and is outside our control. [See this issue](https://github.com/vllm-project/vllm/issues/6145) from the vLLM repo.
   - This is annoying because when users attempt category-specific generation for locally-hosted models (as supported in #512), only the first category result for the first model is generated since the error occurs immediately after.
   - To improve the user experience, we now combine all selected test categories into one task and submit that single task to vLLM, splitting the results afterwards.
   - Note: If multiple locally-hosted models are queued for inference, only the tasks of the first model will complete. Subsequent tasks will still fail due to the cleanup phase error from the first model. Therefore, we recommend running the inference command for one model at a time until vLLM rolls out the fix.

3. **Adding Index to Dataset**:
   - Each test file and possible_answer file now includes an index to help match entries.
  
This PR **will not** affect the leaderboard score.


CharlieJCJ (approved):
LGTM

Fanjia-Yan (approved):
Tested with various setup.

PR #545, opened 2024-07-24
[BFCL] Fix Dataset Pre-Processing for Java and JavaScript Test Category, Part 2

HuanzhiMao:
This PR fixes the function doc pre-processing issues for the Java and JavaScript test categories, following up on PR #538. 

- Some unnecessary steps in the `convert_to_tool` function are removed. These steps should not exist as not every model handler calls the `convert_to_tool` function (for example, the OSS models) and would unfairly benefit the models that use it. To make sure that every model gets the same pre-processed function doc, the pre-processing phase needs to be in the `language_specific_pre_processing` function (which is used by every handler).
- Properly handle the inner element type for nested types.


Fanjia-Yan (approved):
LGTM. Offloaded java/javascript preprocessing from tool preprocessing.

ShishirPatil (approved):

PR #548, opened 2024-07-24
Added new Sequencing datasets: SeqATIS, SeqSNIPS, SeqSGD, SeqMultiWOZ,SeqTopV2

sadhana01:
Added two versions sequencing and sequencing lite with their Licenses.
Added code changes pertaining to the two datasets


ShishirPatil (commented):
Thanks for the PR @sadhana01 

1. There are 10 License files in the PR :) This is one too many. Also BFCL and the entire Gorilla project is already under a License which applies to the entire directory so we don't need these files? 
2. Maybe we can move the data files (jsons) from each category into the `data/` sub-directory? We have organized all the data files into a single directory. The good thing is that the files already have the name `SeqATIS` etc so just moving them there should be good? 
3. ~I'll slowly review each of the rest sequentially since there are quite a few of them. ~ Done reviewing

sadhana01:
@ShishirPatil They are different licenses for each dataset. I had initially put all the data files under the data folder. However I was advised by lawyers to include the license for each dataset which is why I created subfolders for them . We can discuss the best way forward .

PR #559, opened 2024-07-28
[BFCL] Handling rate limits

ShishirPatil:
For BFCL generations of hosted models, this will check for rate limit errors and retry thrice. We can choose to include an exponential back-off as is usually recommended, but I think in the interest of keeping it simple this should perhaps be sufficient?

This does NOT affect the leaderboard values.

### DIFF
--- a/berkeley-function-call-leaderboard/openfunctions_evaluation.py
+++ b/berkeley-function-call-leaderboard/openfunctions_evaluation.py
@@ -133,7 +133,6 @@ def generate_results(args, model_name, test_cases_total):
                         and (
                             e.status_code == 429
                             or e.status_code == 503
-                            or e.status_code == 403
                             or e.status_code == 500
                         )
                     ):
--- a/berkeley-function-call-leaderboard/openfunctions_evaluation.py
+++ b/berkeley-function-call-leaderboard/openfunctions_evaluation.py
@@ -92,6 +92,7 @@ def collect_test_cases(test_filename_total, model_name):
 
 def generate_results(args, model_name, test_cases_total):
     RETRY_LIMIT = 3
+    # 60s for the timer to complete. But often we find that even with 60 there is a conflict. So 65 is a safe no.
     RETRY_DELAY = 65  # Delay in seconds
     
     handler = build_handler(model_name, args.temperature, args.top_p, args.max_tokens)
--- a/berkeley-function-call-leaderboard/openfunctions_evaluation.py
+++ b/berkeley-function-call-leaderboard/openfunctions_evaluation.py
@@ -91,6 +91,7 @@ def collect_test_cases(test_filename_total, model_name):
 
 
 def generate_results(args, model_name, test_cases_total):
+    retry_limit = 3
     handler = build_handler(model_name, args.temperature, args.top_p, args.max_tokens)
 
     if handler.model_style == ModelStyle.OSSMODEL:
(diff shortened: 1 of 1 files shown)


HuanzhiMao (changes requested):
1. `retry_count < retry_limit - 1` should be removed, otherwise it would only retry two times (when `retry_count` is 0 and 1).
2. You can also check the status code for indication of rate limit reached. Such as `429 Too Many Requests`, `403 Forbidden` or `503 Service Unavailable`. Purely reply on the text "rate limit reached" might be inaccurate. For example, the below error message from Anthropic API would not be caught:

> Number of requests has exceeded your rate limit. Please try again later or [contact sales](https://www.anthropic.com/contact-sales) to discuss your options for a rate limit increase.

berkeley-function-call-leaderboard/openfunctions_evaluation.py
```diff
@@ -91,7 +91,9 @@ def collect_test_cases(test_filename_total, model_name):
 
 
 def generate_results(args, model_name, test_cases_total):
-    retry_limit = 3
+    RETRY_LIMIT = 3
+    RETRY_DELAY = 65  # Delay in seconds
```
CharlieJCJ, line 95: Curious where is this number 65 coming from? Add more context here if there are references. 
If this number is arbitrary, let's just put 60 seconds for default 1 minute RETRY_DELAY.

berkeley-function-call-leaderboard/openfunctions_evaluation.py
```diff
@@ -114,9 +117,31 @@ def generate_results(args, model_name, test_cases_total):
...
+                    break  # Success, exit the loop
+                except Exception as e:
+                    # TODO: It might be better to handle the exception in the handler itself rather than a universal catch block here, as each handler use different ways to call the endpoint.
+                    # OpenAI has openai.RateLimitError while Anthropic has anthropic.RateLimitError. It would be more robust in the long run. 
+                    if "rate limit reached" in str(e).lower() or (
+                        hasattr(e, "status_code")
+                        and (
+                            e.status_code == 429
+                            or e.status_code == 503
+                            or e.status_code == 403
```
CharlieJCJ, line 136: For status code 429, 503, 403 check, I've gathered the following references from openai and anthropic:

#### Anthropic:
403 - permission_error: Your API key does not have permission to use the specified resource.
429 - rate_limit_error: Your account has hit a rate limit.
503 - N/A (no support)
500 - api_error: An unexpected error has occurred internal to Anthropic’s systems.
https://docs.anthropic.com/en/api/errors

#### OpenAI: 
403 - Country, region, or territory not supported
429 - Rate limit reached for requests
429 - You exceeded your current quota, please check your plan and billing details
503 - The engine is currently overloaded, please try again later	
500 - The server had an error while processing your request.
https://platform.openai.com/docs/guides/error-codes/api-errors

Here, 403 is related to permission/auth error, not related to rate limit. 
Only 429 is related to rate limit error. 
500, 503 are related to internal server error and server high traffic, which the official documentation recommend to retry (which can be checked by this PR as well to retry).
Other error codes in the documentations (e.g. 401) are not very related. 

**Suggest change**: 
- Remove e.status_code == 403 check
- Add e.status_code == 500 check.

CharlieJCJ (changes requested):

berkeley-function-call-leaderboard/openfunctions_evaluation.py
HuanzhiMao: 60s for the timer to complete. But often we find that even with 60 there is a conflict. So 65 is a safe nos

CharlieJCJ: Ok sounds good, can you add this as a comment in code for reference?

berkeley-function-call-leaderboard/openfunctions_evaluation.py
HuanzhiMao: changed

berkeley-function-call-leaderboard/openfunctions_evaluation.py
HuanzhiMao: sure

HuanzhiMao (approved):

CharlieJCJ (approved):

PR #560, opened 2024-07-28
Update 7_open_functions_v2.html with the right url

ShishirPatil:
The Openfunctions-v2 blog workflow was recommending the `http` endpoint on port `8000`. Since, we migrated to https, the end-point needs to be fixed.
Latest end-point: [https://luigi.millennium.berkeley.edu/v1](https://luigi.millennium.berkeley.edu/v1/chat/completions) 

This PR does not affect the BFCL Leaderboard.

### DIFF
--- a/blogs/4_open_functions.html
+++ b/blogs/4_open_functions.html
@@ -118,7 +118,9 @@
                 <div class="body">
                     <p>Using Gorilla OpenFunctions is straightforward:
                         <ol>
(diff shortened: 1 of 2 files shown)


HuanzhiMao (approved):
Might worth mentioning in the blog that this method requires `pip install openai==0.28`

ShishirPatil:
@HuanzhiMao good point. Done.

PR #532, opened 2024-07-18
Add Salesforce xLAM handler and fix minor issues

zuxin666:
This PR will add ability to inference on Salesforce xLAM function-calling models. They are open-sourced on [Huggingface](https://huggingface.co/collections/Salesforce/xlam-models-65f00e2a0a63bbcd1c2dade4). Therefore, no cost per thousand function calls will be recorded. 

Apart from the `xlam_handler`, this PR also:
- improves the `oss_handler` and the `base_handler` by passing the `dtype` (for vllm) as another argument. This is crucial since we have witnessed some output issues if we enforce our `bfloat16` model to be loaded with `float16`.
- fixed one issue in `berkeley-function-call-leaderboard/apply_function_credential_config.py`. Without the fixing, the saved data with this script will not be loaded properly.

@HuanzhiMao Would you please take a look? Thanks for the great benchmark!

PR #564, opened 2024-07-30
Add NousResearch Hermes models to no cost models list

alonsosilvaallende:
Add NousResearch Hermes models to no cost models list

### DIFF
--- a/berkeley-function-call-leaderboard/eval_checker/eval_runner_helper.py
+++ b/berkeley-function-call-leaderboard/eval_checker/eval_runner_helper.py
@@ -541,6 +541,11 @@ NO_COST_MODELS = [
     "nvidia/nemotron-4-340b-instruct",
     "ibm-granite/granite-20b-functioncalling",
     "THUDM/glm-4-9b-chat",
+    "NousResearch/Hermes-2-Pro-Llama-3-8B",
+    "NousResearch/Hermes-2-Pro-Llama-3-70B",
+    "NousResearch/Hermes-2-Pro-Mistral-7B",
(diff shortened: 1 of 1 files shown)


HuanzhiMao (commented):
Hi @alonsosilvaallende, 
The Hermes model should not be added to the `NO_COST_MODELS` list. 
We use the following formula to estimate the cost and latency for locally-hosted models. image 
Since the Hermes model can be loaded into 8 x V100 machines, it will have the `Cost` and `Latency` fields computed as above.

alonsosilvaallende:
Thank you very much for your explanation @HuanzhiMao 
What does 'no cost' mean? I see that Salesforce/xLAM-7b-fc-r has no cost. Does that mean the cost could not be determined?

HuanzhiMao:
> Thank you very much for your explanation @HuanzhiMao What does 'no cost' mean? I see that Salesforce/xLAM-7b-fc-r has no cost. Does that mean the cost could not be determined?

Yes, because the xLAM model cannot be loaded into 8 x V100 machines (`bfloat16` is not compatible with V100). See [here](https://github.com/ShishirPatil/gorilla/pull/532#issuecomment-2251559211).

alonsosilvaallende:
I understand. The name was misleading me. Last question: Also the functionary small models meetkai/functionary-small-v2.4 cannot be run on 8 x V100 machines?

HuanzhiMao:
> I understand. The name was misleading me. Last question: Also the functionary small models meetkai/functionary-small-v2.4 cannot be run on 8 x V100 machines?

Yes, you are right. `functionary-small` does require `bfloat16`. We will fix it.

PR #574, opened 2024-08-07
[BFCL] Set Model Temperature to 0.001 for All Models

### LINKED ISSUE #500: Set Model Temperature to 0 for Consistent Leaderboard Results
The current model generation script (model_handlers) uses a default temperature of 0.7 for inference. This introduces some degree of randomness into the model output generation, leading to potential variability in the evaluation scores from run to run.
For benchmarking purposes, we should set it to 0 for consistency and reliability of the evaluation results.

HuanzhiMao:
The current model response generation script uses a default temperature of 0.7 for inference. This introduces some degree of randomness into the model output generation, leading to potential variability in the evaluation scores from run to run.
For benchmarking purposes, we set it to 0.001 for consistency and reliability of the evaluation results.

resolves #500 , resolves #562 

This will affect the leaderboard score. We will update it shortly.


CharlieJCJ (approved):
Can put default values in constants file, address in refactor pr later, this PR concerns on enabling functionality, LGTM

HuanzhiMao:
> Can put default values in constants file, address in refactor pr later, this PR concerns on enabling functionality, LGTM

Why should we put it in the constants file? I think that will only make the code more complicated? Cause you will then need to import that global constant in every single model handler.
If you are concerned with the default values for argument `temperature=0.001, top_p=1, max_tokens=1000` when init the model handler, we could technically remove them as they will never be used optionally.

ShishirPatil (approved):

PR #524, opened 2024-07-12
[Leaderboard] Overhaul Leaderboard Table

HuanzhiMao:
The current BFCL leaderboard table is built using basic HTML, which has made it increasingly difficult to add new functionalities. To address this, the leaderboard table is overhauled to use the DataTables library. This change will simplify the process of extending functionalities in the future.

This PR **DOES NOT** change the leaderboard score.

PR #590, opened 2024-08-20
Sticky effects with top row and first three columns on the left while scrolling.

Sharonfish:
Hi. I restructured the code to make sure the top row and left three columns be sticky while scrolling in x, y direction.


ShishirPatil:
This has been one of our earliest requested features. Thanks for the PR @Sharonfish

assets/css/leaderboard.css
```diff
@@ -146,3 +147,55 @@ th:not(.detail-header)  {
...
+    position: sticky;
+    left: 0;
+    z-index: 30;
+    background-color: #f1f8ff;
+  }
+
+  #leaderboard tbody td:nth-child(2),
+  #leaderboard thead .sticky-second-column:nth-child(2) {
+    position: sticky;
+    left: 51px;
```
HuanzhiMao, line 179: Where does the value `51px` come from?

assets/css/leaderboard.css
```diff
@@ -146,3 +147,55 @@ th:not(.detail-header)  {
...
+    z-index: 29;
+  }
+
+  tbody td:nth-child(3),
+  #leaderboard thead .sticky-second-column:nth-child(3) {
+    position: sticky;
+    left: 115px;
```
HuanzhiMao, line 186: Same as above.

assets/css/leaderboard.css
```diff
@@ -1,22 +1,22 @@
...
 }
 
-#leaderboard th,
+/* #leaderboard th,
 #leaderboard td {
     border: 1px solid #ddd;
     padding: 8px;
     text-align: center;
-}
+} */
```
HuanzhiMao, line 11: nit: remove the unnecessary code

assets/css/leaderboard.css
```diff
@@ -146,3 +147,55 @@ th:not(.detail-header)  {
...
+  margin: 0;
+}
+
+#leaderboard thead th{
+    z-index: 100; /* Ensure the header stays above other content */
+    /* border-bottom: 1px solid black;  */
+    background-color: #f1f8ff; /* Background for better visibility */
+}
+
+/* Sticky left columns */
```
HuanzhiMao, line 166: Can we disable the frozen column when it's rendered in mobile view (check for view width)? Because mobile devices are just so narrow, and after the frozen column, users can barely get any useful information from them (see screenshot). 
Frozen header is enabled in both laptop and mobile view (not affected). 
image

HuanzhiMao (changes requested):

HuanzhiMao:
Thanks for the PR @Sharonfish!
This PR will close #431, close #524.

index_live.js
```diff
@@ -927,31 +927,31 @@ function createComparer(columnIndex, isAscending) {
...
+//         rowsArray
+//             .sort(createComparer(columnIndex, isAscending))
+//             .forEach(function (row) {
+//                 tbody.appendChild(row);
+//             });
+//     });
+// });
 
-document.getElementById("rank-col").click(); // Sort by rank by default
+// document.getElementById("rank-col").click(); // Sort by rank by default
```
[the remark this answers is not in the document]
Sharonfish: Yes I'll keep that

assets/css/leaderboard.css
Sharonfish: The 51px is the fixed column width (50px + 1px border), as the column width is constant across different screen sizes based on previous implementation.

assets/css/leaderboard.css
Sharonfish: Same as above

assets/css/leaderboard.css
Sharonfish: Done

assets/css/leaderboard.css
Sharonfish: Yes for sure. I'll make the adjustment.

Sharonfish:
Hi. Thanks for the review from Shishir and Hans. I'll open a new pull request with all changes made.

Sharonfish:
Thanks for the review from Shishir and Hans. I'll open a new pull request with all changes made.

HuanzhiMao:
> Hi. Thanks for the review from Shishir and Hans. I'll open a new pull request with all changes made.

You can just push the changes to the same branch. No need to open a new PR.

Sharonfish:
Sounds good. I'll reopen it.

Sharonfish:
Hi. I made several changes:
1. Change the code based on the comments above
2. I experimented with mobile view sticky effects and I think keeping the rank column can be helpful for viewers as they can identify the row easily with ranking numbers. So for the mobile view, I only keep the left most one (for both expand/non-expand versions).
Let me know if anything needs to be adjusted.

HuanzhiMao:
> Hi. I made several changes:
> 
> 1. Change the code based on the comments above
> 2. I experimented with mobile view sticky effects and I think keeping the rank column can be helpful for viewers as they can identify the row easily with ranking numbers. So for the mobile view, I only keep the left most one (for both expand/non-expand versions).
>    Let me know if anything needs to be adjusted.

In the mobile view, can we freeze the model name column instead of the rank column? Feel like it would be more useful to the user.

Sharonfish:
I agree that the model column will be more informative for users. However, the model column is currently on the third from the left so it's kind of weird if we keep the third one to be sticky while having the 1st, 2nd, and other columns flow as we scroll. If we want to keep the model column sticky, I think we need to rearrange the order a bit so that the model column will switch / replace with the rank column. What do you think?

Fanjia-Yan (approved):
UI looks good to me.

PR #605, opened 2024-08-26
Added python script named as raft_local.py to raft directory to run script completely locally using HF models

himanshushukla12:
There are many users who want to use this `raft.py` script locally, but due to dependency of **OpenAI** it ask for some extra bucks to get it done, so for those like me who have good GPU available can use it locally using `hugging-face` models, I also modified the `README.md` for showing how to use the new script named as `raft_local.py` in same directory.
Modified the `requirements.txt` in raft directory for `torch` related dependencies.


ShishirPatil (approved):
Thank you for the PR @himanshushukla12

PR #510, opened 2024-07-07
Make BFCL User-Friendly and Easy to Extend

devanshamin:
Thank you for open-sourcing BFCL and for your efforts in maintaining it. As I explored the codebase, I noticed some areas for improvement, including duplicate functions, constants, and variables that are inaccessible outside their functions, as well as a lack of abstract classes/methods.

To address these issues, I've begun refactoring the codebase to make it more straightforward to follow, customize, install, and extend. Although this work is still in progress, I wanted to share it with the contributors to seek feedback and gauge interest in reviewing or contributing to the PR. I understand that it’s a significant refactor and may require considerable time and effort, so any assistance or feedback would be greatly appreciated.

### Refactor
- [X] `bfcl` package structure
- [ ] [WIP] `benchmark` and `model_handler`
- [ ] `evaluate` and `eval_checker`
- [ ] Update README
- [ ] Update CHANGELOG

### Test
- [ ] Dependencies
- [ ] `benchmark` -> OSS and Proprietary model handlers
- [ ] `evaluate`

Goal: To make BFCL similar to [lm-evaluation-harness](https://github.com/EleutherAI/lm-evaluation-harness) but for function calling.

Please let me know your thoughts and if you would be interested in reviewing or contributing to this PR.

PR #612, opened 2024-08-28
[BFCL] Refactor Model Handler into OSS and Proprietary Components

devanshamin:
This PR reorganizes the model handler by splitting it into two distinct components: an Open Source (OSS) model handler and a Proprietary model handler. This change is part of a series of updates that address the tasks outlined in issue #510.


HuanzhiMao (approved):
LGTM

berkeley-function-call-leaderboard/.env.example
```diff
@@ -0,0 +1,20 @@
...
+# Provide the API key for the model(s) you intend to use
+OPENAI_API_KEY=sk-XXXXXX
+MISTRAL_API_KEY=
+FIREWORKS_API_KEY=
+ANTHROPIC_API_KEY=
+NVIDIA_API_KEY=nvapi-XXXXXX
+YI_API_KEY=
+
+COHERE_API_KEY=
+USE_COHERE_OPTIMIZATION=False # True/False
```
CharlieJCJ, line 11: Good comment here. In the check,
```python
if os.getenv("USE_COHERE_OPTIMIZATION") == "True":
```
it check if the string is "True" or not. But if the user mistakenly input "true", it will goes to unoptimized version, and current code is silent and doesn't error out and says "true" is an invalid input. 

So, nit on this and one way to do parameter validation is to have "if, elif, else" and raise exception pattern in the if-block.

berkeley-function-call-leaderboard/bfcl/eval_checker/executable_python_function.py
```diff
@@ -1,17 +1,17 @@
-import json
+import os
 import math
 import requests
 from custom_exception import NoAPIKeyError
 import time
 
+# Make sure the env variables are populated
+ENV_VARS = ("GEOCODE_API_KEY", "RAPID_API_KEY", "OMDB_API_KEY", "EXCHANGERATE_API_KEY")
```
CharlieJCJ, line 8: nit, ideally goes to a utils file or constants file

HuanzhiMao: yep. future PRs for constant file.

berkeley-function-call-leaderboard/.env.example
HuanzhiMao: good point

berkeley-function-call-leaderboard/README.md
```diff
@@ -54,13 +62,14 @@ To run the executable test categories, there are 4 API keys to include:
...
 
-    All the Rapid APIs we use have free tier usage. As a result, you need to subscribe to those API providers in order to have the executable test environment setup but it will be free of charge!
+    All the Rapid APIs we use have free tier usage. You need to **subscribe** to those API providers in order to have the executable test environment setup but it will be *free of charge*!
 
 2. Exchange Rate API:https://www.exchangerate-api.com
 3. OMDB API: http://www.omdbapi.com/apikey.aspx
 4. Geocode API: https://geocode.maps.co/
 
 The `apply_function_credential_config.py` will automatically search for dataset files in the default `./data/` directory and replace the placeholder values with the actual API keys.
+After you have filled in the necessary values in the `.env` file, you can run the following command to apply the real API keys to the dataset files.
```
CharlieJCJ, line 72: IMO, if we have `.env`, we should change the wordings for this section in README
```
## Evaluating different models on the BFCL
Make sure the model API keys are included in your environment variables. Running proprietary models like GPTs, Claude, Mistral-X will require them.
```bash
export OPENAI_API_KEY=sk-XXXXXX
export MISTRAL_API_KEY=XXXXXX
export FIRE_WORKS_API_KEY=XXXXXX
export ANTHROPIC_API_KEY=XXXXXX
export COHERE_API_KEY=XXXXXX
export NVIDIA_API_KEY=nvapi-XXXXXX
export YI_API_KEY=XXXXXX


```

And only allow `.env` for passing environment variables for clarity and consistency.

berkeley-function-call-leaderboard/.env.example
HuanzhiMao: resolved.

berkeley-function-call-leaderboard/README.md
HuanzhiMao: resolved

CharlieJCJ (approved):

ShishirPatil (approved):

PR #617, opened 2024-08-29
Update gemini_handler.py to better handle NL+FC model output

vandyxiaowei:
Gemini model is capable of outputting NL/text (for reasoning or verbosity) for the corresponding FC as different parts. Simply concatenation of text with FC will result in decode failures. So we should only extract FC parts for decoding and eval when it exists in the content. 

See below example:

OUTPUT:  ["Okay, I can help you with that. Let's calculate the future value for each investment option:\n\n\n**Bond:**\n\n", {"calculate_future_value": "{\"present_value\": 5000, \"periods\": 10, \"interest_rate\": 0.05}"}, "\n\n**Mutual Fund:**\n\n", {"calculate_future_value": "{\"periods\": 15, \"interest_rate\": 0.07, \"present_value\": 2000}"}, "\n\n\n**Stocks:**\n\n", {"calculate_future_value": "{\"periods\": 20, \"interest_rate\": 0.1, \"present_value\": 1000}"}"]

### DIFF
--- a/berkeley-function-call-leaderboard/bfcl/model_handler/gemini_handler.py
+++ b/berkeley-function-call-leaderboard/bfcl/model_handler/gemini_handler.py
@@ -79,14 +79,15 @@ class GeminiHandler(BaseHandler):
                 "processed_tool": functions,
             }
         try:
-            parts = []
+            fc_parts = []
+            text_parts = []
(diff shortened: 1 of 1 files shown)


HuanzhiMao (approved):
Thanks for the PR!
LGTM

PR #623, opened 2024-09-04
[BFCL] Fix Decoding Issue in Nvidia Handler

HuanzhiMao:
In this PR:

1. Fix decoding issue in the `NvidiaHandler`. An unnecessary whitespace in the decode method caused many model responses fail to get decoded (should be `[` instead of `[ `, and `]` instead of ` ]`).
3. Remove duplicate `ArcticHandler` class. It is the exact same as `NvidiaHandler`, so we just let `snowflake/arctic` use `NvidiaHandler` instead.  

This will affect the leaderboard score for `nvidia/nemotron-4-340b-instruct` and `snowflake/arctic`. We will update it shortly. 

Credit to @mattf for pointing these out. 

---------

Co-authored-by: Matthew Farrellee <mfarrellee@nvidia.com>


Fanjia-Yan (approved):
LGTM

ShishirPatil (approved):

PR #624, opened 2024-09-05
Fix data augmentation examples

az1326:
Fix data augmentation examples in Blog 7: Gorilla OpenFunctions v2.


HuanzhiMao (approved):
Good catch! Thanks for the PR @az1326

PR #634, opened 2024-09-12
[BFCL] Adding UniGPT model handler

BeerTai:
This PR adds model handler for UniGPT2.0 and UniGPT4.0. UniGPT series models are authored by UniSound. I have sent the relevant information of api to  [huanzhimao@berkeley.edu](mailto:huanzhimao@berkeley.edu)


HuanzhiMao:
Apologize for the long delay. Will definitely take a look after the Monday ICLR deadline.

HuanzhiMao (changes requested):
Hi @BeerTai,

Thanks for the PR. 

Any model on the leaderboard needs to be publicly accessible, either open-source or a public API endpoint for inference (you can have login/token, etc, but the general public should be able to eventually get access to that endpoint). Otherwise, you can feel free to benchmark its performance using BFCL, but we won't be able to display that model on the leaderboard unfortunately.

I know you have sent me API keys, but if someone else wants to run the model, how can they obtain such an API Key?

PR #636, opened 2024-09-12
[BFCL] Hot Fix to Remove Extra Parameters for NoAPIKeyError

HuanzhiMao:
This PR won't impact the leaderboard score.


ShishirPatil (commented):
Why are we doing this? Isn't it ok for users to not include all the keys. They may choose to provide only a subset of keys?

CharlieJCJ (approved):

HuanzhiMao:
> Why are we doing this? Isn't it ok for users to not include all the keys. They may choose to provide only a subset of keys?

All 4 executable API keys are required for the `exec` test category. If only a subset is provided, the evaluation result will still be unreliable.

ShishirPatil (approved):

PR #641, opened 2024-09-18
Fix the mean_latency calculation error in the OSS model

zhangch-ss:
hi,


zhangch-ss:
Hello, there may be an anomaly in the mean latency calculation of the OSS model, and I have attempted to fix it.

PR #642, opened 2024-09-18
Add Qwen handler and fix mean_latency calculation error for OSS models

zhangch-ss:
Hello, there may be an anomaly in the mean latency calculation of the OSS model, and I have attempted to fix it.


zhangch-ss:
line 1045, berkeley-function-call-leaderboard/bfcl/eval_checker/eval_runner_helper.py

zhangch-ss:
Qwen handler is also added in this PR
image

zhangch-ss:
Now it supports the newly released Qwen2.5
image

HuanzhiMao:
Apologize for the long delay. Will definitely take a look after the Monday ICLR deadline.

HuanzhiMao (approved):
Thanks for the PR @zhangch-ss and looking forward to the amazing Qwen models. 
I made a few changes to your PR so that the handler is compatible with the latest codebase.
Regarding the mean/latency calculation, I will update those in a separate PR, as that involves quite a bit of change.

HuanzhiMao:
Hey @zhangch-ss ,
I was looking at the chat template for Qwen models (https://huggingface.co/Qwen/Qwen2.5-72B-Instruct/blob/main/tokenizer_config.json) cause I want to add the 72B models to BFCL as well. 
In this PR, Qwen is benchmarked using its normal prompting mode. However I do see that, according to the template, Qwen seems to have a Function Calling mode as well. I'm wondering is there a specific reason why you are not using its FC mode? Would you recommend we add its FC mode as well (in addition to its prompt mode)?

PR #654, opened 2024-09-23
removed unnecessary comments in raft/raft_local.py

himanshushukla12:
Last time when I contributed the `raft_local.py` in directory named `raft` there was some unnecessary were there, which I removed in this pull request. It will not confuse the developers when they read the file.


ShishirPatil (approved):
Thanks for this @himanshushukla12

Issue #652, opened 2024-09-23
[BFCL] Bugs in decode executable predictions


XuHwang:
**Describe the issue**
BFCLv3 has modified the decode logic for executable predictions, while I found there are a lot of cases in REST category that could not be decoded by current method.

For example, here is one example:
```python
result="[requests.get(url=\"https://covid-193.p.rapidapi.com/statistics\", headers={\"X-RapidAPI-Key\": \"xxxxxxx\", \"X-RapidAPI-Host\": \"covid-193.p.rapidapi.com\"}, timeout=5, params={\"country\": \"China\"})]"
```

It would be parsed by [`decoded_output_to_execution_list`](https://github.com/ShishirPatil/gorilla/blob/70b6a4a2144597b1f99d1f4d3185d35d7ee532a4/berkeley-function-call-leaderboard/bfcl/model_handler/utils.py#L767) function. 

Then the nested value `{"X-RapidAPI-Key": "xxxxxxx", "X-RapidAPI-Host": \"covid-193.p.rapidapi.com"}` would  be parsed by [`parse_nested_value`](https://github.com/ShishirPatil/gorilla/blob/70b6a4a2144597b1f99d1f4d3185d35d7ee532a4/berkeley-function-call-leaderboard/bfcl/model_handler/utils.py#L749), which would raise error "`str` object has no attribute `.items()`"

In summary, this logic only consider the case that the nested value is a sub function while ignoring the truth that the argument type is just a dict.

**ID datapoint**
1. Datapoint / Model Handler permalink: 
2. Issue: 
2. Gorilla repo commit #: 

**What is the issue**

**Proposed Changes**

{
 'previous_datapoint':[], 
 'updated_datapoint':[]
}

**Additional context**

PR #660, opened 2024-09-24
[BFCL] Bug Fix parse_nested_value function for model_handler utils

VishnuSuresh27:
In the parse_nested_value function, added a check to determine whether we are dealing with another function call or if its a regular dictionary. Previous version of the code incorrectly assumed that this was always a function call and did not consider the case where the function argument is a dictionary. 

Fix https://github.com/ShishirPatil/gorilla/issues/652

### DIFF
--- a/berkeley-function-call-leaderboard/CHANGELOG.md
+++ b/berkeley-function-call-leaderboard/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 All notable changes to the Berkeley Function Calling Leaderboard will be documented in this file.
 
+- [Sept 25, 2024] [#660](https://github.com/ShishirPatil/gorilla/pull/660): Bug fix in `parse_nested_value` function to handle nested dictionary values properly. 
 - [Sept 19, 2024] [#644](https://github.com/ShishirPatil/gorilla/pull/644): BFCL V3 release:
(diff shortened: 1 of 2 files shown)


HuanzhiMao (changes requested):
Failed on the following test:
```
x = "[foo(a=1, b={'a': 1, 'b': 'hello'}), bar(c=3, d={'a': 1, 'b': foo(c=1, d=2)}), baz(e=f(x=1, y=2))]"

print(default_decode_execute_prompting(x))

# Expect ["foo(a=1, b={'a': 1, 'b': 'hello'})", 'bar(c=3, d={'a': 1, 'b': foo(c=1, d=2)})', 'baz(e=f(x=1, y=2))']
# But got ["foo(a=1, b={a=1, b='hello'})", 'bar(c=3, d={a=1, b=foo(c=1, d=2)})', 'baz(e=f(x=1, y=2))']
```

VishnuSuresh27:
@HuanzhiMao Thank you for catching. Should be fixed in the latest commit.

HuanzhiMao (approved):
LGTM. Thanks @VishnuSuresh27

ShishirPatil (approved):

Issue #643, opened 2024-09-18
[BFCL] Wether there is any semantic checker for the parameters' value？


Ghevil:
**Describe the issue**
When we do the evaluation, do we have some checker that checks the **semantic consistency** of the parameters instead of simply matching them from the candidate list？

It seems too harsh that  some minor errors that lead to errors, such as **San Diego** and **San Diego, CA** are different in format, but the two are semantically consistent.

What's more, like this error "Invalid value for parameter 'items': ['pumpkin', 'egg']. Expected one of [['pumpkins', 'eggs'], ['pumpkin', 'dozen eggs']]."

Looking forward to your reply.

HuanzhiMao:
Hi @Ghevil,

Unfortunately, we don't have any semantic checkers at this moment. 

In addition to trying to make the candidate list as comprehensive as possible, we have also tried to be clear and pose restrictions on the parameter formats in the function documentation. For example, in the location example you provided, we would phrase it as the following so that only `San Diego, CA` is correct and `San Diego` would be wrong. 
```
{
    "location": {
        "type": "string",
        "description": "The location in 'city, state' format."
    }
}
```
As another example, for things like date, there would only be one correct ground truth. 
```
{
    "end_date": {
        "type": "string",
        "description": "The ending date until which to retrieve stock prices. Format: 'yyyy-mm-dd'."
    }
}
```
If you have any other good solutions, we would love to hear your thoughts.

PR #661, opened 2024-09-27
[BFCL] Dataset and Possible Answer Fix

HuanzhiMao:
Total number of entries affected: 8

- Live Simple: 5 entry
  - `live_simple_11-3-7, live_simple_144-95-1, live_simple_160-95-17, live_simple_184-109-0, live_simple_191-115-1`
- Live Multiple: 1 entry
  - `live_multiple_935`
- Parallel Multiple: 2 entry
  - `parallel_multiple_83, parallel_multiple_94`

This will affect the leaderboard score. We will update it in a separate PR. 

We want to thank @budhiraja for pointing out these dataset issues.


berkeley-function-call-leaderboard/data/possible_answer/BFCL_v3_live_multiple.json
```diff
@@ -933,7 +933,7 @@
...
 {"id": "live_multiple_934-191-22", "ground_truth": [{"get_service_providers": {"avg_rating": ["", null], "province_id": [1], "district_name": ["Chatuchak"], "sub_district_name": ["", null], "start_available_date": ["2024-03-22 00:00:00"], "end_available_date": ["2024-03-22 23:59:59"], "min_age": ["", null], "max_age": ["", null], "has_quality_problem": ["", false], "has_late_check_in": ["", false], "is_excellent": [true], "is_package": ["", false], "is_subscription": ["", false], "service_id": [1], "available_for_pet": ["", false], "professional_group_id": ["", null], "job_qty": ["", null], "is_cleaning_condo": ["", false], "is_cleaning_home": ["", false], "is_cleaning_office": ["", false]}}]}
-{"id": "live_multiple_935-192-0", "ground_truth": [{"set_countdown": {"duration": ["01:00"], "purpose": ["meeting"]}}]}
```
budhiraja, line 936: can we do something better here? for example checking that the string has "meeting" rather very specific case-specific strings?

berkeley-function-call-leaderboard/data/possible_answer/BFCL_v3_parallel_multiple.json
```diff
@@ -92,7 +92,7 @@
...
 {"id": "parallel_multiple_92", "ground_truth": [{"calculate.weight_in_space": {"weight_earth_kg": [75.0], "planet": ["Mars"]}}, {"currency_conversion": {"amount": [5000.0], "from_currency": ["USD", "US Dollars", "US Dollar"], "to_currency": ["JPY", "Japanese Yen"]}}, {"unit_conversion.convert": {"value": [24.0], "from_unit": ["in", "inch", "inches"], "to_unit": ["cm", "centimeter", "centimeters"]}}]}
 {"id": "parallel_multiple_93", "ground_truth": [{"geology.get_era": {"era_name": ["Jurassic"], "calculate_years_ago": [true]}}, {"history.get_event_date": {"event_name": ["signing of the Magna Carta", "Magna Carta"], "calculate_years_ago": [true]}}]}
-{"id": "parallel_multiple_94", "ground_truth": [{"sort_list": {"elements": [["apple", "banana", "cherry", "date", "elderberry"], ["elderberry", "cherry", "banana", "apple", "date"]], "order": ["desc", "descending"]}}, {"filter_list": {"elements": [["apple", "banana", "cherry", "date", "elderberry"]], "condition": ["b", "B", "startswith(b)"]}}, {"sum_elements": {"elements": [[5, 10, 15, 20, 25]]}}, {"sort_list": {"elements": [[35, 10, 25, 5, 15]], "order": ["asc", ""]}}]}
```
budhiraja, line 95: why is "b" or "B" considered correct? IMO the prompt says a condition but with "b"/"B" it could as well mean starts with or end with "b"/"B" or something else.

budhiraja (changes requested):

berkeley-function-call-leaderboard/data/possible_answer/BFCL_v3_parallel_multiple.json
HuanzhiMao: Fair point. "b" has been removed from the possible answer list.

berkeley-function-call-leaderboard/data/possible_answer/BFCL_v3_live_multiple.json
HuanzhiMao: As mentioned in #643, unfortunately, we don't have any semantic checkers for the parameter values at this moment. We will explore that option and keep you updated.

budhiraja: Ok, thanks. Suggestion: while evaluating model performance, can we lowercase all param values, given that is what the ground truth is restricted to it seems?

berkeley-function-call-leaderboard/data/possible_answer/BFCL_v3_live_simple.json
```diff
@@ -158,7 +158,7 @@
 {"id": "live_simple_157-95-14", "ground_truth": [{"cmd_controller.execute": {"command": ["echo %time%"], "unit": ["", "N/A"]}}]}
 {"id": "live_simple_158-95-15", "ground_truth": [{"cmd_controller.execute": {"command": ["taskkill /IM code.exe /F", "taskkill /F /IM code.exe"], "unit": ["", "N/A"]}}]}
 {"id": "live_simple_159-95-16", "ground_truth": [{"cmd_controller.execute": {"command": ["timeout 10 && echo \"bring a condom\""], "unit": ["seconds"]}}]}
-{"id": "live_simple_160-95-17", "ground_truth": [{"cmd_controller.execute": {"command": ["start microsoft.windows.camera:"], "unit": ["", "N/A"]}}]}
```
budhiraja, line 161: just to be clarify: if we remove `"unit": ["", "N/A"]`  from ground truth, for test time it wouldn't matter what value the model will put in the arg, right?

berkeley-function-call-leaderboard/data/possible_answer/BFCL_v3_live_multiple.json
HuanzhiMao: Yes. This is the current strategy we use. When the parameter value is a string, we compare the lowercase version of the model response with the lowercase version of the ground truth. Code [here](https://github.com/ShishirPatil/gorilla/blob/main/berkeley-function-call-leaderboard/bfcl/eval_checker/ast_eval/ast_checker.py#L166-L191).

berkeley-function-call-leaderboard/data/possible_answer/BFCL_v3_live_simple.json
HuanzhiMao: In general, if the ground truth doesn't have `"unit": ["", "N/A"]`, then that means `unit` is not one of the valid parameters for that function. If the model provides some value for `unit`, then it will be flagged wrong (unexpected argument). 

For this specific entry, I have updated the func doc as well to remove the `unit` argument, cited below. So it doesn't make sense for the model to provide the `unit` argument now. 
```
{
    "id": "live_simple_160-95-17",
    "question": [[{"role": "user", "content": "open camara"}]],
    "function": [
        {
            "name": "cmd_controller.execute",
            "description": "Executes a system-level command using os.system() on Windows operating systems. It can execute single or multiple commands chained with '&&'.",
            "parameters": {
                "type": "dict",
                "required": ["command"],
                "properties": {
                    "command": {
                        "type": "string",
                        "description": "The system command to be executed. To execute multiple commands, separate them with '&&'. For example, 'dir && echo done'.",
                    }
                },
            },
        }
    ],
}
```

budhiraja: thanks ! there are at least 4-6 more examples where this function is used and has this "unit" param. I dont think the current PR cover these, right? if it doesn't, can we please add this fix in this PR itself?

HuanzhiMao: Yes, we do plan to go through the dataset to fix entries with similar issues before merging this PR. I have sent folks to work on this; it takes a bit of time.

budhiraja:
thanks, here are some more ids where Ground truth seems ambiguous. Can you please take a look? happy to accept the review for this one and follow up in next PR if you prefer that

live_simple_117-73-0, live_simple_33-10-0 [these two are a bit more worrysome in terms of prompts],
id live_simple_50-22-0 [models predicting GB are penalized but GB should be correct because that is what mentioned in prompt],
live_simple_153-95-10 [PuTTY discussion we had offline],
simple_183 [ground truth should allow Santa Clara because that is by default a county? so maybe ground truth is not correct]
simple_375 : location is not required?

CharlieJCJ (approved):
LGTM

budhiraja (approved):

budhiraja:
approving it since no major concerns. please update the changelog (in case its not correct) before merging this PR.

berkeley-function-call-leaderboard/CHANGELOG.md
```diff
@@ -17,6 +17,10 @@ All notable changes to the Berkeley Function Calling Leaderboard will be documen
   - `microsoft/Phi-3-small-8k-instruct`
   - `microsoft/Phi-3-mini-128k-instruct`
   - `microsoft/Phi-3-mini-4k-instruct`
+- [Sept 27, 2024] [#661](https://github.com/ShishirPatil/gorilla/pull/661): Bug fix in the dataset and possible answers.
```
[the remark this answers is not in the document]
HuanzhiMao: Yea we are waiting for one more change to be landed. Will update the change log after.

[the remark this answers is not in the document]
HuanzhiMao: updated

HuanzhiMao:
This PR mainly addresses issues with the Live dataset. The change log has been updated with the latest numbers.

PR #667, opened 2024-09-30
update hammer handler and  add Hammer2.0 model

linqq9:
Hello, we have updated the hammer handle and added Hammer2.0 series models, including [Hammer2.0-7b](https://huggingface.co/MadeAgents/Hammer2.0-7b), [Hammer2.0-3b](https://huggingface.co/MadeAgents/Hammer2.0-3b), [Hammer2.0-1.5b](https://huggingface.co/MadeAgents/Hammer2.0-1.5b) and [Hammer2.0-0.5b](https://huggingface.co/MadeAgents/Hammer2.0-0.5b). The performance on BFCL-V3 is as follows:

| Model                          | Overall Acc | Non-live AST | Non-live Exec | Live AST | Multi Turn Acc | Relevance | Irrelevance |
|--------------------------------|-------------|--------------|---------------|----------|----------------|-----------|-------------|
| MadeAgents/Hammer2.0-7b (FC)   | 56.60       | 90.15        | 82.64         | 68.68    | 15.75          | 92.68     | 68.20       |
| MadeAgents/Hammer2.0-1.5b (FC) | 51.94       | 84.31        | 81.80         | 63.17    | 11.38          | 92.68     | 61.83       |
| MadeAgents/Hammer2.0-3b (FC)   | 49.88       | 86.77        | 80.25         | 66.06    | 0.50           | 92.68     | 68.59       |
| MadeAgents/Hammer2.0-0.5b (FC) | 39.51       | 67.00        | 65.73         | 51.62    | 0.00           | 87.80     | 67.00       |


HuanzhiMao (changes requested):
Hey @linqq9, Thanks for the PR!

Question regarding the `_format_prompt` function: 

1. In [these lines](https://github.com/ShishirPatil/gorilla/pull/667/files#diff-908811121f031427f4d79ccddce8ff4529fb194f77b70223f7b85b5dd6fe97f2R95-L76), why we remove all those `<|im_start|>` `<|im_end|>` tags? The chat template for Hammer2.0 on huggingface does include them. 
2. What happened to the system prompts [here](https://github.com/ShishirPatil/gorilla/pull/667/files#diff-908811121f031427f4d79ccddce8ff4529fb194f77b70223f7b85b5dd6fe97f2R69-R87)? Previously they were contacted in the `task instruction` section, but now it seems that they are all thrown away.
3. Why do we need to special handle the situation when the length of the prompt is 2 (code [here](https://github.com/ShishirPatil/gorilla/pull/667/files#diff-908811121f031427f4d79ccddce8ff4529fb194f77b70223f7b85b5dd6fe97f2R70-R74))?

linqq9:
> Hey @linqq9, Thanks for the PR!
> 
> Question regarding the `_format_prompt` function:
> 
> 1. In [these lines](https://github.com/ShishirPatil/gorilla/pull/667/files#diff-908811121f031427f4d79ccddce8ff4529fb194f77b70223f7b85b5dd6fe97f2R95-L76), why we remove all those `<|im_start|>` `<|im_end|>` tags? The chat template for Hammer2.0 on huggingface does include them.
> 2. What happened to the system prompts [here](https://github.com/ShishirPatil/gorilla/pull/667/files#diff-908811121f031427f4d79ccddce8ff4529fb194f77b70223f7b85b5dd6fe97f2R69-R87)? Previously they were contacted in the `task instruction` section, but now it seems that they are all thrown away.
> 3. Why do we need to special handle the situation when the length of the prompt is 2 (code [here](https://github.com/ShishirPatil/gorilla/pull/667/files#diff-908811121f031427f4d79ccddce8ff4529fb194f77b70223f7b85b5dd6fe97f2R70-R74))?

@HuanzhiMao hi，Thanks for your questions. Here are my responses:
Regarding the <|im_start|> <|im_end|> tags, in _query_prompting, we use the client.chat.completions.create function. This function already includes handling of these special symbols for the user's provided content.
As for the system prompts, the content in the system is later included in the generated content. Hence, there is no need to add it redundantly and it is removed.
We use the length of the prompt being 2 to determine if historical information is included. When the length of the prompt is 2 (system and user), it is considered that there is no historical information. If it is greater than 2, it means historical information is present.

linqq9:
Hi @HuanzhiMao, Do you have any other questions?

HuanzhiMao:
I will submit a PR to your branch.

linqq9:
> I will submit a PR to your branch.

ok, thanks!

HuanzhiMao:
Hi @linqq9, I have submitted a PR to your branch. https://github.com/MadeAgents/gorilla/pull/2
Things I changed:

1. Let `HammerHandler` inherit from `OSSHandler` instead and copy over any necessary decoding logic from the `SalesforceHandler`. This simplifies things a lot.
2. Change to use `completions` endpoint instead of `chat.completions`. This won't affect the score. 
3. Since Hammer doesn't take user-supplied system message, we turn any system message into user message (only message role change). This affects the live categories, and is also what we do to other models in similar situations. 
4. Since Hammer has its own system prompt, we don't need to add the default BFCL system prompt in `_pre_query_processing_prompting`. 

ps, after the change, the score is roughly the same as you reported for `MadeAgents/Hammer2.0-7b (FC)`.

PR #644, opened 2024-09-19
[BFCL] BFCL v3

HuanzhiMao:
BFCL V3 release. Introducing new multi-turn dataset and state-based evaluation metric for category: `multi_turn_base`, `multi_turn_miss_func`, `multi_turn_miss_param`, `multi_turn_long_context`, `multi_turn_composite`;  a significant leap towards multi-turn, and multi-step function calling (tool usage) benchmarking.

BFCL V3 is a critical advancement in evaluating how Large Language Models (LLMs) interact with diverse scenarios through invoking right functions. Multi-turn function calling allows models to engage in a back-and-forth interaction with users, making it possible for LLMs to navigate through the complex tasks by asking clarifying questions. In contrast to multi-turn `(user t0, assistant t1, user t2, assistant t3, ..)`, multi-step is where the LLM can break the response down into multiple steps `(user t0, assistant t1, assistant t2,..)`. This new paradigm mimics real-world agentic behaviors where AI assistants might have to plan execution paths, request and extract critical information, and handle sequential function invokes to complete a task.

To read more about the composition and construction of this live dataset, please refer to our [blog](https://gorilla.cs.berkeley.edu/blogs/13_bfcl_v3_multi_turn.html). 

---------

**Also in this PR**:

1. Switch to use vllm serve for OSS model inference
2. Switch to Vertex AI Python SDK for Gemini models inference
3. Split out ast_checker and executable_checker for readability
4. Several outdated or deprecated models will be excluded from the leaderboard and replaced with their updated successors to improve the leaderboard's overall maintainability.

---------

Co-authored-by: Fanjia Yan <fanjiayan@berkeley.edu>
Co-authored-by: Charlie Cheng-Jie Ji <charliechengjieji@berkeley.edu>
Co-authored-by: Jason Huang <jasonhuang1103@berkeley.edu>
Co-authored-by: Vishnu Suresh <vishnusuresh@berkeley.edu>
Co-authored-by: Yixin Huang <yixinhuang1@berkeley.edu>
Co-authored-by: Xiaowen Yu <yxw2002@berkeley.edu>

PR #671, opened 2024-10-04
[BFCL] Speed Up Locally-hosted Model Inference Process

### LINKED ISSUE #649: [BFCL] The generation in v3 is too slow
**Describe the issue**
A clear and concise description of what the issue is.

**ID datapoint**
1. Datapoint / Model Handler permalink: 
2. Issue: 
2. Gorilla repo commit #: 

**What is the issue**

After migration to version 3, where the vllm API serving strategy is adopted, the generation speed shows a severe degradation. It takes about 6x more time to get all results. 

**Proposed Changes**

{
 'previous_datapoint':[], 
 'updated_datapoint':[]
}

**Additional context**

HuanzhiMao:
Fix #649 

Instead of send requests to the vllm server one by one in sequence, we should send all requests all at once to vllm to utiliza its batching and optimizaiton benefits. 

Tested on 8 x A100 (40G) with Llama 3.1 70B. The inference speed on single-turn entries are roughtly the same (within 1 minute difference) as when using `llm.generate` before the BFCL V3 release in #644]. The multi-turn entries still takes around 2 hours to complete, but that's largely due to the nature of the multi-turn dataset; it has been much faster than previously where it would take 2 days to finish.

This PR **will not** affect the leaderboard score.

### DIFF
--- a/berkeley-function-call-leaderboard/CHANGELOG.md
+++ b/berkeley-function-call-leaderboard/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 All notable changes to the Berkeley Function Calling Leaderboard will be documented in this file.
 
+- [Oct 4, 2024] [#671](https://github.com/ShishirPatil/gorilla/pull/671): Speed up locally-hosted model's inference process by parallelizing the inference requests.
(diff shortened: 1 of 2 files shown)


CharlieJCJ (approved):
LGTM

PR #679, opened 2024-10-10
[BFCL-v3]  Multi-Turn Possible Answer Order Change

Fanjia-Yan:
This PR fix the possible answer misalignment for `miss_func` and `composite`. The dataset entry will encounter a question with no suitable function, resulting in model chatting and then the dataset will supplement the suitable function, resulting in a function call.

Previous:
```
[["cd(folder='workspace')", "ls(a=True)"], [], ["mv(source='log.txt',destination='archive')"], ["cd(folder='archive')", "grep(file_name='log.txt',pattern='Error')"], ["tail(file_name='log.txt',lines=20)"]]
```
Current:
```
[[], ["cd(folder='workspace')", "ls(a=True)"], ["mv(source='log.txt',destination='archive')"], ["cd(folder='archive')", "grep(file_name='log.txt',pattern='Error')"], ["tail(file_name='log.txt',lines=20)"]]
```

The empty list signifies that no function should be invoked. No function should be invoked one turn earlier when no relevant functions are found.


HuanzhiMao (approved):

CharlieJCJ (approved):
Tested, LGTM

PR #681, opened 2024-10-11
[BFCL Blog] BFCL V3 wording refine

CharlieJCJ:
Changes: 
- Be clear that BFCL V3 consist of existing datasets from BFCL V1 and V2. BFCL V3 introduces a new category, i.e. multi-turn, multi-step into the mix. 
- Route users to earlier blog posts if they are not familiar with the concept of function calling, and evaluations in single-turn single-step scenarios. These knowledge are foundational and prerequisite for BFCL V3 blog.


HuanzhiMao (approved):
LGTM.

PR #691, opened 2024-10-13
added FitnessTracker API Collection for multi-turn for BFCL

amitojsingh2022:
The collection I am designing is the FitnessTrackerAPI which focuses on helping users manage their fitness and nutrition goals. It includes methods which allow users to log physical activities, track calories burned, food intake including calories and protein, along with tracking the daily fitness goals. This API can be used to have a chain of very interesting function calls incorporating the other APIs.

I am a new UC Berkeley third-year studying EECS and Business looking to get involved with this research and received an email recently to contribute and make a PR by tonight! Handling this high-priority task from the excel sheet: Add multi-turn executable function source code for BFCL

### DIFF
--- a/berkeley-function-call-leaderboard/bfcl/eval_checker/multi_turn_eval/func_source_code/fitness_tracker.py
+++ b/berkeley-function-call-leaderboard/bfcl/eval_checker/multi_turn_eval/func_source_code/fitness_tracker.py
@@ -1,0 +1,158 @@
+from typing import Dict, List, Union, Optional
+from datetime import datetime
+
+class FitnessTrackerAPI:
+    """
+    A class representing the Fitness Tracker API for managing fitness activities.
+
+    Attributes:
+        activities (List[Dict[str, Union[str, float]]]): List of logged activities which include name, and calories burned
+        food_log (List[Dict[str, Union[str, float]]]): List of logged food intake include calories and protein eaten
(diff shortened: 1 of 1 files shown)


HuanzhiMao (approved):
Thanks for the PR @amitojsingh2022 and welcome!

Fanjia-Yan (changes requested):
Hi Amitoj,

Thank you for the contribution and we would like to work with you to further polishing this API set so that it can be incorporated into our evaluations! Here are 2 improvements on the particular `FitnessTrackerAPI`:

1. It would be cool if some output of the functions can serve as inputs of a new function. Is it possible we can add more of that in? One example is "Eat apple until I fulfill my goal". A user need to find the remaining goal, and use the output as an input to somehwere like batch eating apple.
2. It would be interesting if we create an Enum class of {food : calories} and a way to populate this Enum via functions. This can promote some interesting questions like "Eat apple until I fulfill my goal". That way the model need to constantly check if he meets the goal. 

Thank you for pushing this through and let us know if you have any question.

amitojsingh2022:
Quick summary of the improvements from the last commit: Instead of creating an enum class for the food since enums are immutable. food_log is a dictionary which stores the food item's name, calories, and protein information which I use for functions to complete tasks such as "Eat an apple until I fulfill my goal". In addition, added some new functions to take on more dynamic tasks.

PR #645, opened 2024-09-20
[BFCL] Add Blog 13: BFCL V3: Introducing Multi-Turn & Multi-Step Function Calling

CharlieJCJ:
This PR adds the blog for BFCL V3: Introducing Multi-Turn & Multi-Step Function Calling

---------

Co-authored-by: Fanjia Yan <fanjiayan@berkeley.edu>
Co-authored-by: Yixin Huang <yixinhuang1@berkeley.edu>
Co-authored-by: Jason Huang <jasonhuang1103@berkeley.edu>
Co-authored-by: Vishnu Suresh <vishnusuresh@berkeley.edu>
Co-authored-by: Xiaowen Yu <yxw2002@berkeley.edu>
Co-authored-by: Huanzhi Mao <huanzhimao@berkeley.edu>

PR #692, opened 2024-10-14
[BFCL Blog] Changes BFCL V3 Images

Raymond112514:
Joint work by 
Shreya: [shreyam04@berkeley.edu](url)
Raymond: [r112358@berkeley.edu](url)

* Changed multistep/multiturn definition image.
* Changed the dataset composition image to tree diagram to visualize the relationship better. 
* Added example for multi turn dataset.


CharlieJCJ:
Thanks @Raymond112514 for the PR, looks great! 

Some feedback to the images:
1. For [this](https://github.com/ShishirPatil/gorilla/pull/692/files#diff-f79490e0ccd73ffee5575803e9b71a9e0743ffd3c436df7a627050476dfa3a85), can you make the substitute BFCL V3 with BFCL V3 • Multi-Turn & Multi-Step? since BFCL V3 as a whole represents 
image
2. For [this](https://github.com/ShishirPatil/gorilla/pull/692/files#diff-e1b326a6960c59a027030c7226bc366b613b27e8021738d60b4e5386e490216c), can you make the background orange, so that visually it's consistent for representing multi-turn. 
3. In [this](https://github.com/ShishirPatil/gorilla/pull/692/files#diff-e1b326a6960c59a027030c7226bc366b613b27e8021738d60b4e5386e490216c), somewhere you would need to mention `Gorilla Fiile System` as the API class that's being used, so that when user sees the Initial config, they would have more context. 

There are a couple additional wording changes in the image annotation, I'll take a closer look and suggest them tmr. Here are some style changes that you can make first. 

Also, please add Shreya's github handle in the PR description :) , similar to the PR description listed in #645 as an example. 

Thanks again.

PR #696, opened 2024-10-14
[BFCL] Add Gemma-2 models

jacovkim:
BFCL currently don't support gemma2-2b. Added handler to handler map using GemmaHandler.


HuanzhiMao:
Does the old Gemma handler still work for Gemma 2?

jacovkim:
Yes, the format of the prompt is same across two versions.

HuanzhiMao (approved):
Thanks for the PR @jacovkim and welcome! 
LGTM

PR #700, opened 2024-10-15
Continued Groundwork for GoEx APIs

anay-20:
Worked on Spotify api, using Spotipy.
Created more functions for Spotify API


goex/localhost-key.pem
ShishirPatil, line 1: Should this be public and commited to the repository?

ShishirPatil (commented):
Thank you for the PR of adding spotify! Do you mind also appending to the README on how folks can use this along with GOEX? If you have a link to a video of screen-grab demonstrating it, that would also be nice.

PR #701, opened 2024-10-15
Fixing BFCL-v3 multi-turn apps

virginie-do:
Fixed MessageAPI by adding self._random as a field (causing issues with ground truth function calls provided in multi_turn_base_40 and multi_turn_base_43 + their augmentations)

Fixed TravelAPI by adding RMB as a currency (causing issues with ground truth function calls in 'multi_turn_base_160', 'multi_turn_base_162', 'multi_turn_base_164', 'multi_turn_base_167' + their augmentations)


ShishirPatil:
Thanks for the PR @virginie-do 
@HuanzhiMao  can you just look at `self._random: random.Random = random.Random(200191)`? Rest of it looks good to me.

berkeley-function-call-leaderboard/bfcl/eval_checker/multi_turn_eval/func_source_code/travel_booking.py
```diff
@@ -570,7 +589,7 @@ def compute_exchange_rate(
 
         """
         exchange_rates = {
-            ("USD", "RMS"): 7,
+            ("USD", "RMB"): 7,
```
HuanzhiMao, line 592: Should be `RMS` here.

ShishirPatil: RMB is a currency, but not RMS? So, maybe the right thing to do here is RMB?

HuanzhiMao: [Here](https://github.com/ShishirPatil/gorilla/blob/956fa0df849c425ade127bdd63bd51e599790831/berkeley-function-call-leaderboard/bfcl/eval_checker/multi_turn_eval/func_source_code/travel_booking.py#L672), `RMS` is a 'fake' airport code that stands for `Rivermist`. The inclusion of fake airport is to avoid model retrieving the airport code from its memory instead of using the relative functions.

berkeley-function-call-leaderboard/bfcl/eval_checker/multi_turn_eval/func_source_code/message_api.py
```diff
@@ -67,6 +71,7 @@ def __init__(self):
         }
         self.message_count: int = 0  # useless(?)
         self.current_user: Optional[str] = None
+        self._random: random.Random = random.Random(200191)
```
HuanzhiMao, line 74: We already have a `self._random` initialized in the `_load_scenario` method, why do we need to do it in the `__init__` again?

berkeley-function-call-leaderboard/bfcl/eval_checker/multi_turn_eval/func_source_code/travel_booking.py
virginie-do: It seems from the variable and function name that RMB should be a currency. If you believe it should still be RMS then it would be good to fix the ground truths that call this function with 'RMB'. I listed a few above.

berkeley-function-call-leaderboard/bfcl/eval_checker/multi_turn_eval/func_source_code/message_api.py
virginie-do: It raises an error when calling the function generate_id() (in send_message)

berkeley-function-call-leaderboard/bfcl/eval_checker/multi_turn_eval/func_source_code/travel_booking.py
ShishirPatil: This should be RMB. Resolved offline.

ShishirPatil (approved):

HuanzhiMao:
Hi @virginie-do,
Regarding the `RMS` issue, it is an oversight on my end. I misread the variable name and it should indeed be `RMB` instead. I apologize. We use `RMS` for airport name and `RMB` for currency. 
Regarding the `_random` thing, our team cannot reproduce it on my end. I executed the ground truth in order for `multi_turn_base_40` and it gives no error. So I will revert that change. If you could provide the error log, we can open a new issue for that.

```
MessageAPI_instance = MessageAPI()
MessageAPI_instance._load_scenario({})
MessageAPI_instance.message_login(user_id='USR001')
print(MessageAPI_instance.send_message(sender_id='USR001', receiver_id='USR002', message='The report has been finalized.'))
```

terminal output `{'sent_status': True, 'message_id': 67410, 'message': "Message sent to 'USR002' successfully."}`

HuanzhiMao (approved):

PR #712, opened 2024-10-19
[BFCL] Support Dynamic max_tokens for Locally-Hosted Models

HuanzhiMao:
This is an optimization that can sometimes improve the performance of models with low context windows. 

According to OpenAI's spec, the `max_tokens` for the completions endpoint is the maximum number of tokens that can be generated. It does not include the input token count. For example, if a model has a context length of 4096, our input message takes 1000 tokens and you set the `max_tokens` to 4096, then that would error because the total number of tokens (1000 in input and 4096 requesed for output) exceeds the model's context window length. So what we want to do is that, before call the completions endpoint, use the model's tokenizer to figure out how many tokens the input message `formatted_prompt` has used, subtract that amount from the model's maximum context length and supply that value for the `max_token` argument so that we can try to avoid getting the maximum length exceeded error. In short, we want to allow the model to generate as many as possible till the limit of its context length. 

PS, even for the multi turn long context category, in a single turn, the maximum token used by the ground truth function call is much much less than 4096 (calculated using tiktoken). So this should be a safe threshold.

> max: 1299
min: 24
average: 125.96


yutongxie58:
I've implemented the max_tokens calculation based on input token count and specific model limits for various OSS models. Please let me know if any further tweaks are required!

Fanjia-Yan (approved):
LGTM

CharlieJCJ (approved):
LGTM

PR #716, opened 2024-10-23
Add batch processing for ClaudeHandler

yutongxie58:
Added batch processing for all the queries to the Claude models using Anthropic Message Batches API


HuanzhiMao:
Closed by accident. Apologies.

PR #717, opened 2024-10-23
[BFCL] Chore: Separate out Func Doc for Multi-Turn Categories

HuanzhiMao:
This PR separates out the function doc for the multi-turn categories from the dataset file itself. This helps view the changes; otherwise, the git diff would load forever. 
This PR will not affect the leaderboard score.


CharlieJCJ (approved):
LGTM

PR #719, opened 2024-10-24
[BFCL] Multi Turn Dataset and Possible Answer Fix (Base Category)

HuanzhiMao:
This PR addresses a few issues with the base multi turn entries:

1. Some initial config values are in the wrong format, and would result in execution error even when running the ground truth function calls in order.
2. Some entries in the initial config are not used by the `_load_scenario` method.
3. Some ground truth function calls have wrong parameter values, and would result in `{"error": "xxx"}` when executed.
4. Some functions are unreasonable in a real-life setting with its parameter or its functionality.

After this PR, all info in the initial config are used, and the ground truth function calls will not results in any error when executed in order.

This is part of the effort to thoroughly bug fix the multi turn categories. We will have more PR coming in the next few days. 

---------

Co-authored-by: Charlie Cheng-Jie Ji <55744150+CharlieJCJ@users.noreply.github.com>
Co-authored-by: Fanjia-Yan <78303449+Fanjia-Yan@users.noreply.github.com>
Co-authored-by: VishnuSuresh27 <112032533+VishnuSuresh27@users.noreply.github.com>


CharlieJCJ (approved):
Suggested changes in Message API. 

**LGTM**, clear from dataset validation tests.

PR #720, opened 2024-10-24
[BFCL] Adding New Model GoGoAgent

RogueTensor:
GoGoAgent.ai provides an API endpoint that is supported by the Bittensor AI ecosystem.
We don't (nor do we have plans to) charge for use at the moment and users can login to get an API key.
I'll share an API key in discord soon.


HuanzhiMao (approved):
Thanks for the PR @RogueTensor !
LGTM

PR #723, opened 2024-10-29
[BFCL] Multi Turn Dataset Fix (Base Category)

HuanzhiMao:
This PR fixes the ambiguous prompt issue and some wrong ground truth issues for the multi_turn_base category. After this PR, the multi_turn_base entries should be bug-free.

Following #719 and #722 , this is also part of the effort to thoroughly bug fix the multi turn categories. We will have more PR coming in the next few days. 

---------

Co-authored-by: Charlie Cheng-Jie Ji <55744150+CharlieJCJ@users.noreply.github.com>
Co-authored-by: Fanjia-Yan <78303449+Fanjia-Yan@users.noreply.github.com>
Co-authored-by: VishnuSuresh27 <112032533+VishnuSuresh27@users.noreply.github.com>

PR #724, opened 2024-10-29
[BFCL] Multi Turn Pipeline Robustness Patch

HuanzhiMao:
This PR adds checks to prevent unsafe function calls from being run and potentially breaking the evaluation pipeline.

1. In GorillaFileSytem, the `xargs` method is removed. This will not affect the dataset entries, and that function is not used in any entries. 
2. In `execute_multi_turn_func_call`, an input filter has been implemented to prevent the execution of unsafe function calls like `kill`, `exit`, `remove`, etc.


Fanjia-Yan (approved):
LGTM. Corresponding function JSON string is removed #723

CharlieJCJ (approved):
LGTM

PR #730, opened 2024-10-31
[BFCL] Patch #724

HuanzhiMao:
This PR fixes an issue introduced in #724. We need to preserve the original `func_call` to avoid errors during evaluation.
`func_call` should be like `xxxx_instance.cd(folder='document')` so that when `eval(func_call)`, the `cd` method is invoked. Without the patch, `func_call` would become `cd` and thus resulting in variable not found error.

### DIFF
--- a/berkeley-function-call-leaderboard/bfcl/eval_checker/multi_turn_eval/multi_turn_utils.py
+++ b/berkeley-function-call-leaderboard/bfcl/eval_checker/multi_turn_eval/multi_turn_utils.py
@@ -77,16 +77,18 @@ def execute_multi_turn_func_call(
 
         # Evaluate the function call
         try:
+            # We need to make a copy here because otherwise the `eval(func_call)` would error. 
+            func_call_copy = func_call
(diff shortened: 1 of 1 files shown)


berkeley-function-call-leaderboard/bfcl/eval_checker/multi_turn_eval/multi_turn_utils.py
```diff
@@ -77,16 +77,18 @@ def execute_multi_turn_func_call(
 
         # Evaluate the function call
         try:
+            # We need to make a copy here because otherwise the `eval(func_call)` would error. 
+            func_call_copy = func_call
```
budhiraja, line 81: shouldn't this be copy.deepcoopy(func_call) ?

budhiraja: actually nevermind, strings are good

PR #731, opened 2024-10-31
[BFCL] Small typo in variable name in travel_booking.py

daanaea:
self.token_type was mistyped


HuanzhiMao (approved):
Good catch! Thanks for the PR @daanaea !

PR #736, opened 2024-11-04
Add the option to log to WandB during bfcl evaluate

ThomasRochefortB:
In our internal use of BFCL at Valence Labs, we use WandB extensively to centralize the results of our benchmarking.
This PR adds a `--wandb-project` CLI argument to the `bfcl evaluate` command to upload the generated `.csv` files to wandb.

Here is an example command:
```bash
bfcl evaluate --model gpt-3.5-turbo-0125  
--wandb-project <wandb_entity:wandb_project>
```

This will log the `data_live.csv`, `data_non_live.csv` and `data_overall.csv` as dataframe artifacts on the WandB `wandb_project` under the `wandb_entity`.

### DIFF
--- a/berkeley-function-call-leaderboard/README.md
+++ b/berkeley-function-call-leaderboard/README.md
@@ -309,6 +309,12 @@ If you want to run `live_simple` and `javascript` tests for a few models and `go
 bfcl evaluate --model gorilla-openfunctions-v2 claude-3-5-sonnet-20240620 gpt-4-0125-preview gemini-1.5-pro-preview-0514 --test-category live_simple javascript
 ```
 
+If you want to log the evaluation results as WandB artifacts to a specific WandB entity and project, you can use the following argument:
+```bash
+bfcl evaluate --model gpt-3.5-turbo-0125  
+--wandb-project <wandb_entity:wandb_project>
+```
+
 ### Model-Specific Optimization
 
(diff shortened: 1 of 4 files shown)


HuanzhiMao (commented):
Thanks for the PR @ThomasRochefortB !

Two suggestions:
1. We should add `wandb` to the `pyproject.toml`; otherwise the pipeline will break. Maybe as an optional-dependency since `wandb` is not required for evaluation to complete?
2. Maybe we should add the wandb_project to an environment variable (in the `.env` file) and just use a CLI flag to signal that the result should be added to the wandb because it's likely user would use the same wandb_project across runs. So they don't need to type it in the command line everytime, and it makes the code cleaner. 

What do you think?

ThomasRochefortB:
Thank you for the feedback @HuanzhiMao !

I think these are 2 great suggestions. I will adjust my branch accordingly.

HuanzhiMao:
> Thank you for the feedback @HuanzhiMao !
> 
> I think these are 2 great suggestions. I will adjust my branch accordingly.

Sounds good!

ThomasRochefortB:
There you go @HuanzhiMao ! I have made the following changes:
- Added an optional env var `WANDB_BFCL_PROJECT`  in the .env.example
- Modified the `init_wandb` function to read from the env var
- Added Wandb as an optional dependencies on the pyoroject.toml

HuanzhiMao (approved):
LGTM
Thanks for the PR @ThomasRochefortB !

PR #743, opened 2024-11-08
[BFCL] Add Support for Regeneration, Specific Test Entry IDs, and Custom Directory Locations

Raymond112514:
This pull request introduces several improvements and new features to the CLI commands:

1. **Regeneration Support**  
   - Added the `--allow-overwrite` flag to the generation command.  
   - This option allows regeneration of test entries even if some entries already exist.  
   - The flag is only valid for the `generate` command.

2. **Selective Test Entry Execution**  
   - Introduced a new `--run-ids` flag:  
     - When enabled, this argument reads a list of test entry IDs from the file `test_case_ids_to_generate.json`.  
     - Only those specific test IDs will be executed, instead of the entire category.  
     - This feature is also exclusive to the `generate` command, and cannot be used together with `--test-category`.

3. **Customizable Result and Score Directories**  
   - Added `--score-dir` and `--result-dir` options for both the `generate` and `evaluate` commands.  
   - These options allow users to specify custom paths for result and score directories.  
   - Paths should be relative to the root folder of `berkeley-function-call-leaderboard`.

In addition, this PR contains an update to the `check_illegal_python_param_name.py` script to avoid storing the functions in the multi_turn categories; it won't affect the accuracy of the dataset.


Raymond112514:
Added the `--rerun-all` flag. When this flag is present, the results are overwritten. 
Changed the logic of `collect_test_case` slightly.

Raymond112514:
Added the `--result-dir` and `--score dir` option.

HuanzhiMao (approved):
Thanks for the PR @Raymond112514! I made a few modifications to simplify the code logic and fixed a few bugs. Also there are a few places where you forget to change. Feel free to take a look at my commits.
I have updated the PR description as well. 

LGTM

CharlieJCJ:
Testing on my side: 
image
image

CharlieJCJ (approved):
LGTM

PR #746, opened 2024-11-08
[BFCL] Chore: Improve Inference Log Readability

HuanzhiMao:
This PR improves the readability of the inference log; makes it read more like a conversation-style. The instruction on how to interpret/understand the log is also included. The inference log is included as part of the result file.

This will not change the leaderboard score.


CharlieJCJ (approved):
LGTM

PR #751, opened 2024-11-11
[BFCL] Prompt Caching for Claude Models

VishnuSuresh27:
This PR request seeks to merge my changes of adding prompt caching abilities when running inference on Claude models. The benefit will be reduced cost significantly for inference on BFCL's multi-turn datasets when using the following models (in both Function Calling and Prompting modes):
- Claude 3.5 Sonnet
- Claude 3 Haiku
- Claude 3 Opus

Summary of changes made:

- Cached user messages
- Cached system prompt (for Prompting mode)
- Cached tools (for Function-Calling mode)

Please note:

- This implementation rightfully avoids caching in single-turn cases as there aren't any future turns that could avail cache reading benefits.
- According to the [Anthropic guide](https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching#cache-storage-and-sharing), using prompting caching **will not** affect the model accuracy.
  > Prompt caching has no effect on output token generation. The response you receive will be identical to what you would get if prompt caching was not used.


berkeley-function-call-leaderboard/bfcl/model_handler/proprietary_model/claude.py
```diff
@@ -129,17 +147,25 @@ def _parse_query_response_FC(self, api_response: any) -> dict:
...
             "input_token": api_response.usage.input_tokens,
             "output_token": api_response.usage.output_tokens,
+            "prompt_write_cache_token_count": getattr(api_response.usage, 'cache_creation_input_tokens', 0),
+            "prompt_read_cache_token_count": getattr(api_response.usage, 'cache_read_input_tokens', 0)
         }
 
     def add_first_turn_message_FC(
         self, inference_data: dict, first_turn_message: list[dict]
     ) -> dict:
+        if isinstance(first_turn_message[0]['content'], str):
```
HuanzhiMao, line 157: This should apply for all the messages, not just the first one. Because this function is also called in single turn scenarios.

HuanzhiMao (approved):
LGTM. 
According to the [Anthropic guide](https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching#cache-storage-and-sharing), using prompting caching will not affect the model accuracy.
> Prompt caching has no effect on output token generation. The response you receive will be identical to what you would get if prompt caching was not used.

PR #755, opened 2024-11-12
[BFCL] Add support for Writer models and Palmyra X 004

samjulien:
This PR adds support for Writer models and our latest Palmyra X 004 to BFCL. Thank you!

### DIFF
--- a/berkeley-function-call-leaderboard/.env.example
+++ b/berkeley-function-call-leaderboard/.env.example
@@ -7,6 +7,7 @@ ANTHROPIC_API_KEY=
 NVIDIA_API_KEY=nvapi-XXXXXX
 YI_API_KEY=
 GOGOAGENT_API_KEY=
+WRITER_API_KEY=
 
 # We use Vertex AI to inference Google Gemini models
 VERTEX_AI_PROJECT_ID=
--- a/berkeley-function-call-leaderboard/CHANGELOG.md
+++ b/berkeley-function-call-leaderboard/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 All notable changes to the Berkeley Function Calling Leaderboard will be documented in this file.
 
+- [Nov 26, 2024] [#755](https://github.com/ShishirPatil/gorilla/pull/755): Add new model `palmyra-x-004` to the leaderboard.
 - [Nov 25, 2024] [#718](https://github.com/ShishirPatil/gorilla/pull/718): Add new model `openbmb/MiniCPM3-4B-FC` to the leaderboard.
 - [Nov 25, 2024] [#697](https://github.com/ShishirPatil/gorilla/pull/697): Add the following new models to the leaderboard:
   - `deepseek-ai/DeepSeek-V2.5`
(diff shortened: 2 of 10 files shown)


HuanzhiMao (commented):
Thanks for the PR @samjulien !

I noticed that [here](https://github.com/ShishirPatil/gorilla/pull/755/files#diff-f91567bd5a8ad1ec1409cf88da95fdcdb63a3d56f0ed3b34a36bb862e6bd06b7R52-R55), you intentionally removed all the response field from the function doc. 
For model that doesn't support `response` field, we add those information to the end of the function description because we don't want information loss (code [here](https://github.com/ShishirPatil/gorilla/blob/main/berkeley-function-call-leaderboard/bfcl/model_handler/utils.py#L218-L228)). Is there a reason why you chose to do it that way?

samjulien:
Hi @HuanzhiMao, I made that change you suggested and rebased the branch to be updated with `main`, but I'm getting some weird errors now that I wasn't getting before these updates. Looks like it is happening in `multi_threaded_inference`, getting `KeyError: 'function'`. I'll keep investigating, but maybe you can take a look or have seen this with other providers? Thanks!

HuanzhiMao:
Hey @samjulien, I have submitted a PR to your fork. 
Two things I changed:
1. All the methods ending with `_prompting` are removed, because they are never used. Palmyra is a Funciton Calling model, and so only those `_FC` methods will be used. 
2. I made the WriterHandler inherit from the OpenAIHandler since they share the same logic for decoding, tool handling, etc. The only difference is `_query_FC` because writerai use `self.client.chat.chat` instead of `self.client.chat.completions.create`
I have tested the code locally and it is working perfectly fine.

HuanzhiMao:
> Thank you so much @HuanzhiMao, great thinking! I really appreciate you doing that. I've merged it in.
> 
> One other question: We don't usually append `-FC` to `palmyra-x-004` since we don't have a separate non-FC model, but it seemed like it is necessary for the tests to run correctly. Am I right on that?

Yes. The `-FC` is mainly to signal the BFCL user that this is a function calling model (instead of a prompt model). You can see that when we call the endpoint, we removed the `-FC` in the model name:
> api_response = self.client.chat.chat(
                messages=message,
                model=self.model_name.replace("-FC", ""),
                temperature=self.temperature,
                tools=tools,
                tool_choice="auto",
            )

HuanzhiMao (approved):
LGTM. Thanks for the PR!

HuanzhiMao:
Update: I removed the `_FC` in the model name.

samjulien:
Thank you so much @HuanzhiMao! Let me know if you need an API key for the leaderboard.

HuanzhiMao:
> Thank you so much @HuanzhiMao! Let me know if you need an API key for the leaderboard.

DM-ed via Discord.

Gdewilde:
Nice addition. Why is this model not visible on the leaderboard? 
Screenshot 2025-01-30 at 23 57 23

PR #756, opened 2024-11-13
introducing custom handlers

sangmandu:
Hi there,

I am a current user of bfcl and would like to submit this PR to address an improvement opportunity I've identified while working with the module.

Currently, model handlers are managed through the `handler_map.py`, which presents some challenges. This managed approach requires code updates from maintainers whenever new models are released. Additionally, users must wait for updates if their desired model isn't yet supported.

While technically proficient users might implement their own solutions for unsupported models, this especially becomes problematic when using bfcl as a submodule. Modifying the original code is not ideal from a maintenance perspective, as it requires additional resources for ongoing maintenance.

Therefore, I'm submitting this PR to enhance the main codebase.

The current `build_handler` function only allows fetching predefined model handlers. I propose implementing a `HandlerLoader` that enables users to input custom handlers. This approach minimizes changes to the main codebase while allowing users to manage their custom handlers through just two files: a Python file containing the handler definition and a JSON configuration file.

### DIFF
--- a/berkeley-function-call-leaderboard/bfcl/__main__.py
+++ b/berkeley-function-call-leaderboard/bfcl/__main__.py
@@ -2,6 +2,8 @@ import csv
 from collections import namedtuple
 from datetime import datetime
 from typing import List
+import json
+import os
 
 import typer
 from bfcl._llm_response_generation import main as generation_main
@@ -10,6 +12,7 @@ from bfcl.eval_checker import eval_runner
 from bfcl.model_handler.handler_map import HANDLER_MAP
 from dotenv import load_dotenv
 from tabulate import tabulate
+from bfcl.model_handler.handler_loader import HandlerLoader
 
 
 class ExecutionOrderGroup(typer.core.TyperGroup):
@@ -52,8 +55,20 @@ def models():
     """
     List available models.
     """
+    available_models = set(HANDLER_MAP.keys())
+
+    # If a custom handler setting exists, add it to the
+    handler_config_path = os.getenv("BFCL_HANDLER_CONFIG")
+    if handler_config_path and os.path.exists(handler_config_path):
+        try:
+            with open(handler_config_path) as f:
+                handler_config = json.load(f)
+                available_models.update(handler_config.keys())
+        except Exception as e:
+            print(f"Warning: Error loading custom handler config: {str(e)}")
+
     table = tabulate(
-        [[model] for model in HANDLER_MAP.keys()],
+        [[model] for model in sorted(available_models)],
         tablefmt="plain",
         colalign=("left",),
     )
--- a/berkeley-function-call-leaderboard/bfcl/_llm_response_generation.py
+++ b/berkeley-function-call-leaderboard/bfcl/_llm_response_generation.py
@@ -15,7 +15,7 @@ from bfcl.constant import (
     TEST_FILE_MAPPING,
 )
 from bfcl.eval_checker.eval_runner_helper import load_file
-from bfcl.model_handler.handler_map import HANDLER_MAP
+from bfcl.model_handler.handler_loader import HandlerLoader
 from bfcl.model_handler.model_style import ModelStyle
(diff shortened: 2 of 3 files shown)


berkeley-function-call-leaderboard/bfcl/__main__.py
```diff
@@ -52,8 +55,20 @@ def models():
...
+    # If a custom handler setting exists, add it to the
+    handler_config_path = os.getenv("BFCL_HANDLER_CONFIG")
+    if handler_config_path and os.path.exists(handler_config_path):
+        try:
+            with open(handler_config_path) as f:
+                handler_config = json.load(f)
+                available_models.update(handler_config.keys())
+        except Exception as e:
+            print(f"Warning: Error loading custom handler config: {str(e)}")
+
```
sangmandu, line 69: Previously, you could simply use dictionary mapping to get handler.
For now,  HandlerLoader will controll.

You can set environment variable for `BFCL_HANDLER_CONFIG`.
This environ var. point to the path of json file.
and if this exists, Loader can find the path of json file.

sangmandu: This is example of json files
```json
{
    "my-custom-model": {
        "module_path": "/path/to/my_handlers.py",
        "class_name": "MyCustomHandler"
    }
}
```

and example of setting env var.
```
# you can enter this on command"
export BFCL_HANDLER_CONFIG=/path/to/handler_config.json

python -m bfcl generate --model my-custom-model --test-category all
```

HuanzhiMao (commented):
Thanks for the PR @sangmandu ! Could you provide an example `BFCL_HANDLER_CONFIG`?

berkeley-function-call-leaderboard/bfcl/__main__.py
sangmandu: At `/path/to/my_handlers.py` you can make custom handlers clas
```python
from bfcl.model_handler.base_handler import BaseHandler
from bfcl.model_handler.model_style import ModelStyle

class MyCustomHandler(BaseHandler):
    def __init__(self, model_name, temperature):
        super().__init__(model_name, temperature)
        self.model_style = ...
        ...
```

sangmandu:
> Thanks for the PR @sangmandu ! Could you provide an example `BFCL_HANDLER_CONFIG`?

@HuanzhiMao 

Wow, rapid reply! Thanks.
It looks like our timing is a little off, and I've provided an example in the code section.

berkeley-function-call-leaderboard/bfcl/eval_checker/eval_runner_helper.py
```diff
@@ -244,6 +245,9 @@ def get_cost_letency_info(model_name, cost_data, latency_data):
 def generate_leaderboard_csv(
     leaderboard_table, output_path, eval_models=None, eval_categories=None
 ):
+    # Load metadata at the beginning of the function
+    model_metadata, _, _ = metadata_loader.load_metadata()
+
```
sangmandu, line 250: For dynamic load, we'd used `model_metadata` rather than `MODEL_METADATA_MAPPING`

berkeley-function-call-leaderboard/bfcl/eval_checker/metadata_loader.py
```diff
@@ -0,0 +1,54 @@
...
+            - metadata: Dict mapping model names to their metadata
+            - prices: Dict mapping model names to their prices
+            - no_cost_models: List of model names that have no associated cost
+        """
+        metadata = dict(MODEL_METADATA_MAPPING)
+        prices = dict(OUTPUT_PRICE_PER_MILLION_TOKEN)
+        no_cost = list(NO_COST_MODELS)
+
+        # Check for additional metadata config file path in environment variables
+        metadata_config_path = os.getenv("BFCL_MODEL_METADATA")
```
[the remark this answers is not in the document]
sangmandu: The result after evaluation

image

PR #760, opened 2024-11-14
Update google-cloud-aiplatform version to 1.72.0

gabrielibagon:
Recently the google-cloud-aiplatform library has been updated to version 1.72.0, which fixes typecasting bugs in Function Calling. This affects many cases in BFCL, so it would be great to use this version of the library for an accurate score of the model quality.

The changelog describing the bugfix is here: https://github.com/googleapis/python-aiplatform/blob/main/CHANGELOG.md

Thanks!


HuanzhiMao (approved):
Thanks for the PR @gabrielibagon !
LGTM

PR #658, opened 2024-09-24
[BFCL] Add gemini-1.5-pro-002, gemini-1.5-pro-002-FC, gemini-1.5-pro-001, gemini-1.5-pro-001-FC, gemini-1.5-flash-002, gemini-1.5-flash-002-FC, gemini-1.0-pro-002, gemini-1.0-pro-002-FC

HuanzhiMao:
This PR adds the following new models to the leaderboard:

  - `gemini-1.5-pro-002`
  - `gemini-1.5-pro-002-FC`
  - `gemini-1.5-pro-001`
  - `gemini-1.5-pro-001-FC`
  - `gemini-1.5-flash-002`
  - `gemini-1.5-flash-002-FC`
  - `gemini-1.0-pro-002`
  - `gemini-1.0-pro-002-FC`

Note: All the code for Gemini prompting series inference are already written out in #644, but they were not thoroughly verified at that time due to limited bandwidth. This PR has the code tested and thus we 'officially' add support for the prompting version of the Gemini models.

PR #761, opened 2024-11-15
[BFCL] Remove Workaround Patch for Vertex AI Package

HuanzhiMao:
Since the typecasting bugs in Gemini Function Calling has been fixed in the upstream  `google-cloud-aiplatform` library and the version has been updated in #760 , the workaround patch for the Vertex AI package introduced in #658 is no longer needed and should be removed for code readability.


CharlieJCJ (approved):
LGTM

PR #762, opened 2024-11-15
[BFCL Chore] Supply `data_multi_turn.csv` for Multi-Turn Evaluation Results

HuanzhiMao:
Similar to `data_live.csv` and `data_non_live.csv`, we also provide a `data_multi_turn.csv` for the evaluation results of only multi-turn categories.


CharlieJCJ (approved):
LGTM

PR #768, opened 2024-11-17
Skip adding empty content from gemini

gabrielibagon:
This PR fixes errors where the user adds an empty turn in a multi-turn conversation request, causing the following API error message:

```
InvalidArgument: 400 Please use a valid role: user, model.
```

Gemini sometimes returns an empty content during multi-turn conversations, particularly in response to a function result in the previous turn. 

As a workaround, we can just skip adding this turn and proceed with the next user turn, which should not affect the correctness of the example unless the missing turn was a function call. But generally, it seems that the missing turn is the optional natural language response to the function response.

### DIFF
--- a/berkeley-function-call-leaderboard/CHANGELOG.md
+++ b/berkeley-function-call-leaderboard/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 All notable changes to the Berkeley Function Calling Leaderboard will be documented in this file.
 
+- [Nov 17, 2024] [#767](https://github.com/ShishirPatil/gorilla/pull/767): Fix price and latency calculation. A merge conflict results in a duplicate line, and counting the input and output token for each entry multiple times.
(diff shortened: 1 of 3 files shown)


HuanzhiMao (commented):
Hey @gabrielibagon ,
Thanks for the PR.
I think it might be better if we handle it similar to how we do it in the prompting mode? 
Because if the model didn't output any response, then in your implementation, the chat history would be like `user, model, tool, user, model...`, which is not ideal because each `tool` message should have a `model` response before the next `user` query to make it complete; it should be like `user, model, tool, model, user, model`. I will submit a PR to your branch with my proposed implementation. Let me know what you think.

gabrielibagon:
@HuanzhiMao That approach looks good, and I agree that it is good to align with the prompting strategy. I've merged your addition to this PR. Thank you!

HuanzhiMao (approved):
LGTM

PR #769, opened 2024-11-17
Remove stop condition in gemini retry logic

gabrielibagon:
Even with the current retry logic, I was still facing ResourceExhaustion errors during the multi_turn_long_context cases. I have removed the stop condition in the retry logic to allow the code to continue trying until the tokens per minute quota allows for new requests. This potentially could increase the duration of the test, but it will allow all examples to finish.


HuanzhiMao (approved):
Thanks for the PR @gabrielibagon !

PR #775, opened 2024-11-22
add transformers to pyproject.toml

gabrielibagon:
After installing the repo:

```
cd berkeley-function-call-leaderboard/
python3 -m pip install -e .
```

When trying to run on OSS models, I run into the following error -

```
ModuleNotFoundError: No module named 'transformers'
```

This PR adds transformers library into the pyproject.toml

### DIFF
--- a/berkeley-function-call-leaderboard/pyproject.toml
+++ b/berkeley-function-call-leaderboard/pyproject.toml
@@ -30,7 +30,8 @@ dependencies = [
     "tabulate>=0.9.0",
     "google-cloud-aiplatform==1.72.0",
     "mpmath==1.3.0",
-    "tenacity==9.0.0"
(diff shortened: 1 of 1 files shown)


HuanzhiMao:
Hey @gabrielibagon ,
As per [setup instruction](https://github.com/ShishirPatil/gorilla/tree/main/berkeley-function-call-leaderboard#install-dependencies), when running the OSS models, you should use either `pip install -e .[oss_eval_vllm]` or `pip install -e .[oss_eval_sglang]`.

gabrielibagon:
Ah that makes sense. This resolves the issue. Thank you @HuanzhiMao!

gabrielibagon:
Ah that makes sense. This resolves the issue. Thank you @HuanzhiMao!

PR #781, opened 2024-11-23
[BFCL Chore] Add Retry Mechanism with Backoff for Rate Limit Handling Across Proprietary Models

HuanzhiMao:
This PR introduces a retry mechanism with backoff for handling rate limit errors and resource exhaustion across multiple proprietary models. A `retry_with_backoff` decorator is added so that we can just apply this decorator to the relevant methods in each model handlers.
Currently, Claude, Gemini, and OpenAI have the retry logic enabled, because user are most likely to hit rate limit issues with these three endpoints. Mistral doesn't have a specific rate limit error defined in their Python SDK.


CharlieJCJ (approved):
@Fanjia-Yan LGTM

PR #805, opened 2024-11-30
[BFCL] Minor Fix to Claude Handler for Better Prompt Caching

HuanzhiMao:
This PR updates the `_add_execution_results_prompting` method in Claude handler to have the `content` field in the message be a list of one dictionary `"content": [{"type": "text", "text": "abcd"}]` instead of a string `"content": "abcd"`. Prompt caching expects it in this format.


CharlieJCJ (approved):
LGTM

PR #808, opened 2024-12-02
[BFCL] Improve Latency Measurement Accuracy and Enable Default State Logging

HuanzhiMao:
This pull request improves the accuracy of latency measurement by addressing the following issues:

1. **Retry Latency Handling:** Previously, when a retry occurred, the reported latency included the duration from the first attempt to the final successful attempt, resulting in inflated latency values. This change ensures that only the latency of the final successful attempt is measured. This adjustment primarily impacts the Gemini and Claude models.

2. **Preprocessing Time Exclusion:** For models requiring preprocessing before hitting their endpoints, this preprocessing time is now excluded from the reported latency, ensuring a more accurate reflection of the actual request duration.

Additionally, the `state_log` is now included as part of the `inference_log` by default. Users can disable it if needed by using the `--exclude-state-log` flag.

### DIFF
--- a/berkeley-function-call-leaderboard/CHANGELOG.md
+++ b/berkeley-function-call-leaderboard/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 All notable changes to the Berkeley Function Calling Leaderboard will be documented in this file.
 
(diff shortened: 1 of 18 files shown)


Fanjia-Yan (approved):
LGTM

PR #809, opened 2024-12-03
[BFCL] Resolve Issue in Gemini Model When No Model Output

HuanzhiMao:
When the Gemini models output no content, the `api_response.candidates` will be `[]`, and so calling `api_response.candidates[0].content.parts` will give `IndexError: list index out of range`. We should check if there is a candidate before accessing its content.


HuanzhiMao:
Hey @gabrielibagon, could you kindly verify if this is the best way to patch it?

gabrielibagon:
This looks good to me. I'm curious on what situations Gemini returns no candidates vs no parts, since I've not often encountered no candidates being returned. But this seems like the right approach for that situation.

HuanzhiMao:
> This looks good to me. I'm curious on what situations Gemini returns no candidates vs no parts, since I've not often encountered no candidates being returned. But this seems like the right approach for that situation.

The previous implementation of checking `len(api_response.candidates[0].content.parts)` works most of the time. It is only in the irrelevance category that sometimes there are no candidates. And this happens pretty often in the prompt mode (such as `gemini-1.5-pro-002`).

PR #828, opened 2024-12-11
[BFCL Chore] Add Unit Test for Valid Func Doc Format

HuanzhiMao:
This PR adds a unit test to check for the correct function doc format. For example, a required parameter should not have a 'default' field.

Below is the output, which indicates potential issues that need to be addressed (they will be fixed in separate PRs).

```
# python check_func_doc_format.py 
--------------------
Entry ID: exec_multiple_45
Function: book_room
Error: The parameter 'room_type' is of type 'dict' and should contain either 'properties' or 'additionalProperties' field to specify the sub-parameters details.
...
```


Fanjia-Yan (changes requested):
- Remove Commented Code.
- Add instruction to run it on PR description
- List out all possible error cases that could be caught either in codebase or PR description
Other than that LGTM

Fanjia-Yan (approved):
LGTM

PR #837, opened 2024-12-17
[BFCL] Add New Model `Llama-3.3-70B-Instruct`, `Llama-3.3-70B-Instruct-FC`

HuanzhiMao:

  - `meta-llama/Llama-3.3-70B-Instruct-FC`
  - `meta-llama/Llama-3.3-70B-Instruct`

From the [model card](https://www.llama.com/docs/model-cards-and-prompt-formats/llama3_3#-prompt-template-):

> Llama 3.3 uses the same prompt format as Llama 3.1. Prompts written for Llama 3.1 work unchanged with Llama 3.3.
> Llama 3.3 also supports the same code-interpreter and tool-calling capabilities as Llama 3.1.

So we have llama 3.3 use the same handler as llama 3.1 series.

### DIFF
--- a/berkeley-function-call-leaderboard/CHANGELOG.md
+++ b/berkeley-function-call-leaderboard/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 All notable changes to the Berkeley Function Calling Leaderboard will be documented in this file.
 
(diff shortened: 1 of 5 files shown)


Fanjia-Yan (approved):
LGTM

PR #840, opened 2024-12-18
[BFCL] Add `o1-2024-12-17` and `o1-2024-12-17-FC`

HuanzhiMao:
Update `o1-preview-2024-09-12` to `o1-2024-12-17`. Also, since o1 now supports function calling, it's FC mode `o1-2024-12-17-FC` is also added. 
Note that o1-mini still doesn't support function calling.


Fanjia-Yan (approved):
LGTM

PR #848, opened 2024-12-20
[BFCL] Skip Executable Categories When API Keys Missing

HuanzhiMao:
This PR improves the behavior of the generation and evaluation pipeline. When executable categories are involved and API keys are not provided in the `.env` file, instead of throwing an error, the affected categories will now be skipped. This enhancement provides a smoother experience for first-time users.

1. What will happen to overall score? What would be the difference between score on BFCL official leaderboard vs. without Executable?

If the API Key is not provided, that category will not be evaluated and will be treated as 0 by default in the overall score calculation, which means the overall score (and the one on the leaderboard) will be hurt if the API Keys are not supplied. PR #849 should make things more clear. 

2. What percentage of executable are there?

310 in total, out of 4751 entries.


Fanjia-Yan (approved):
Code Change LGTM. Please add in PR description:

1. What will happen to overall score? What would be the difference between score on BFCL official leaderboard vs. without Executable?
2. What percentage of executable are there?

HuanzhiMao:
> Code Change LGTM. Please add in PR description:
> 
> 1. What will happen to overall score? What would be the difference between score on BFCL official leaderboard vs. without Executable?
> 2. What percentage of executable are there?

Updated.

PR #854, opened 2024-12-27
[BFCL] Fix UNDERSCORE_TO_DOT for Locally-Hosted Models 

HuanzhiMao:
For locally hosted models, when doing inference, the `.`  in the function name is replaced with `_` (code section [here](https://github.com/ShishirPatil/gorilla/blob/main/berkeley-function-call-leaderboard/bfcl/model_handler/utils.py#L64-L74)). As a result, these models must be added to the UNDERSCORE_TO_DOT list so that their function names are correctly matched during evaluation. Previously, not all locally hosted models were included in that list. This PR fixed this issue by having the `UNDERSCORE_TO_DOT` list contain all models in the `local_inference_handler_map`.


minpeter:
I have a question before merging this. Why not convert dot to underscore for all models?

PR #855, opened 2024-12-29
[BFCL] Add Mistral Local Serving Handler and Add New Model `mistralai/Ministral-8B-Instruct-2410`

HuanzhiMao:
This PR adds new model `mistralai/Ministral-8B-Instruct-2410` to the leaderboard. Since this model is open-sourced on huggingface, a local-serving mistral handler is also added; this is different from the `mistral.py` in `proprietary_model` folder, which does inference through their API endpoint.

### DIFF
--- a/berkeley-function-call-leaderboard/CHANGELOG.md
+++ b/berkeley-function-call-leaderboard/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 All notable changes to the Berkeley Function Calling Leaderboard will be documented in this file.
 
+- [Dec 29, 2024] [#855](https://github.com/ShishirPatil/gorilla/pull/855): Add new model `mistralai/Ministral-8B-Instruct-2410` to the leaderboard.
 - [Dec 22, 2024] [#838](https://github.com/ShishirPatil/gorilla/pull/838): Fix parameter type mismatch error in possible answers.
   - Simple: 2 affected
   - Multiple: 1 affected
--- a/berkeley-function-call-leaderboard/SUPPORTED_MODELS.md
+++ b/berkeley-function-call-leaderboard/SUPPORTED_MODELS.md
@@ -64,6 +64,7 @@ Below is a comprehensive table of models supported for running leaderboard evalu
 |palmyra-x-004 | Function Calling|
 |BitAgent/GoGoAgent | Prompt|
 |google/gemma-2-{2b,9b,27b}-it 💻| Prompt|
+|mistralai/Ministral-8B-Instruct-2410 💻| Function Calling|
 |meta-llama/Meta-Llama-3-{8B,70B}-Instruct 💻| Prompt|
 |meta-llama/Llama-3.1-{8B,70B}-Instruct-FC 💻| Function Calling|
 |meta-llama/Llama-3.1-{8B,70B}-Instruct 💻| Prompt|
--- a/berkeley-function-call-leaderboard/bfcl/eval_checker/model_metadata.py
+++ b/berkeley-function-call-leaderboard/bfcl/eval_checker/model_metadata.py
@@ -157,6 +157,12 @@ MODEL_METADATA_MAPPING = {
         "Mistral AI",
         "Proprietary",
     ],
+    "mistralai/Ministral-8B-Instruct-2410": [
+        "Ministral-8B-Instruct-2410 (FC)",
+        "https://huggingface.co/mistralai/Ministral-8B-Instruct-2410",
+        "Mistral AI",
+        "Mistral AI Research License",
+    ],
     "claude-3-sonnet-20240229-FC": [
         "Claude-3-Sonnet-20240229 (FC)",
         "https://www.anthropic.com/news/claude-3-family",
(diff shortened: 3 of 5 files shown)


CharlieJCJ (approved):
LGTM

kirill-fedyanin:
Hi, @HuanzhiMao, thanks for adding the model. I initially having some trouble, may be i don't do it correctly because i don't understand the framework so well? so i running
```sh
bfcl generate --test-category=simple --model="mistralai/Ministral-8B-Instruct-2410" --num-gpus=4
bfcl evaluate --test-category=simple --model="mistralai/Ministral-8B-Instruct-2410"
```
and getting accuracy 0

What i notice in the score raw is meaningful, but decoded is emptpy, meaning handler doesn't parse the answer
```json
{ "model_result_raw": "[TOOL_CALLS][{\"name\": \"calculate_triangle_area\", \"arguments\": {\"base\": 10, \"height\": 5}}]",
  "model_result_decoded": [],}
```

so i trace back to decode_ast function, which expected to have list but it getting the raw string

```python
    def decode_ast(self, result, language="Python"):
        # The input is already a list of dictionaries, so no need to decode
        # `[{func1:{param1:val1,...}},{func2:{param2:val2,...}}]`
         if type(result) != list:
            return []
        return result
 ```

i explicetly parsed the strings and accuracy became 86%.

```python
    def decode_ast(self, result, language="Python"):
        # The input is already a list of dictionaries, so no need to decode
        # `[{func1:{param1:val1,...}},{func2:{param2:val2,...}}]`
        if type(result) == str:
            result = result.replace("[TOOL_CALLS]", "")
            function_calls = eval(result)
            decoded_output = []
            for func_call in function_calls:
                name = func_call["name"]
                params = func_call["arguments"]
                decoded_output.append({name: params})

            return decoded_output

        if type(result) != list:
            return []
        return result
```

So, it did work, but doesn't seem like a cleanes of solutions. I wonder why it was initally waiting an array and what could be a reason it didn't work this way for me initally?

HuanzhiMao:
> Hi @HuanzhiMao, thanks for adding the model. I was initially having some trouble, maybe because I don't understand the framework very well? I was running:
> 
> ```shell
> bfcl generate --test-category=simple --model="mistralai/Ministral-8B-Instruct-2410" --num-gpus=4
> bfcl evaluate --test-category=simple --model="mistralai/Ministral-8B-Instruct-2410"
> ```
> 
> and getting accuracy 0
> 
> What I noticed in the score was that the raw output was meaningful, but decoded was empty, meaning the handler doesn't parse the answer:
> 
> ```json
> { "model_result_raw": "[TOOL_CALLS][{\"name\": \"calculate_triangle_area\", \"arguments\": {\"base\": 10, \"height\": 5}}]",
>   "model_result_decoded": []}
> ```
> 
> So I traced back to the decode_ast function, which expected to have a list but was getting the raw string:
> 
> ```python
>     def decode_ast(self, result, language="Python"):
>         # The input is already a list of dictionaries, so no need to decode
>         # `[{func1:{param1:val1,...}},{func2:{param2:val2,...}}]`
>          if type(result) != list:
>             return []
>         return result
> ```
> 
> I explicitly parsed the strings and accuracy became 86%:
> 
> ```python
>     def decode_ast(self, result, language="Python"):
>         # The input is already a list of dictionaries, so no need to decode
>         # `[{func1:{param1:val1,...}},{func2:{param2:val2,...}}]`
>         if type(result) == str:
>             result = result.replace("[TOOL_CALLS]", "")
>             function_calls = eval(result)
>             decoded_output = []
>             for func_call in function_calls:
>                 name = func_call["name"]
>                 params = func_call["arguments"]
>                 decoded_output.append({name: params})
> 
>             return decoded_output
> 
>         if type(result) != list:
>             return []
>         return result
> ```
> 
> So, it did work, but doesn't seem like the cleanest solution. I wonder why it was initially waiting for a list and can i make it behave like that by itself?

Hey @kirill-fedyanin,

I'm not sure why the raw model output on your end contains the `[TOOL_CALLS]` stuff, and I suspect that is what makes it unparsable (using the current handler implementation). In the result/score we obtained (you can find it [here](https://github.com/HuanzhiMao/BFCL-Result/blob/main/2024-12-29/score/mistralai_Ministral-8B-Instruct-2410/BFCL_v3_simple_score.json)), the raw model output never contains the `[TOOL_CALLS]`, and so the handler is working fine and the mistral model is able to get 93.5% correct for the `simple` category. Maybe you could open an issue and attach the result file and score file to help us debug the issue?

To answer your question:
The generation pipeline works like this ([code link](https://github.com/ShishirPatil/gorilla/blob/main/berkeley-function-call-leaderboard/bfcl/model_handler/base_handler.py#L571-L597))
```
inference_data: dict = self._pre_query_processing_prompting(test_entry)
inference_data = self.add_first_turn_message_prompting(
    inference_data, test_entry["question"][0]
)

api_response, query_latency = self._query_prompting(inference_data)

# Try parsing the model response
model_response_data = self._parse_query_response_prompting(api_response)
```
Notice how that after we get the API response from the vllm/sglang server, we feed it to the `_parse_query_response_prompting` function to extract and further process it to get the "real" model response. The processed model response is then treated as the "model_response_raw" and recorded in the result files. 
When you run the evaluation command, the checker will read in the result file (which is the processed model response from `_parse_query_response_prompting`) and feed it to the `decode_ast` or `decode_exec` function to convert it to the correct format that our AST/Executable checker takes in. 
In this case, if the `_parse_query_response_prompting` is able to successfully parse, it would happen to output in the same format that `decode_ast` should output (and thus, if the input to `decode_ast` is a list of dictionaries, `decode_ast` can just return the input without any processing). When it failed to parse, it would output the string (the output from vllm/sglang server).  

Feel free to let me know if you have more questions!

PR #857, opened 2024-12-29
[BFCL] Add New Model `DeepSeek-V3`

HuanzhiMao:
This PR adds new model `DeepSeek-V3` to the leaderboard. 
Even though this model is open-sourced on huggingface, due to its huge model size (685B), we will use its hosted API endpoint for inference.


CharlieJCJ (approved):
LGTM

PR #870, opened 2025-01-05
[BFCL] Fix `live_simple_44-18-0` and `live_simple_45-18-1`

pkesseli:
For these `ThinQ_Connect` tasks, the parameter name in the specification is `currentJobMode`, not `airConJobMode` as expected in the ground truth.


HuanzhiMao (approved):
Great catch! Thanks for fixing that.

PR #871, opened 2025-01-06
[BFCL] Fix live_simple_165-98-0

pkesseli:
`schema` is not marked as required, and the default value is the expected value. Thus we should also allow for `""`.


HuanzhiMao (approved):
Thanks for spotting that! Good fix.

PR #872, opened 2025-01-06
[BFCL] Fix `live_simple_183-108-0`

pkesseli:
Mismatch between function description and ground truth: function description only contains `avg_rating`, ground truth expects `rating`.


HuanzhiMao (approved):
Thanks for the fix! Nice catch.

PR #881, opened 2025-01-11
[BFCL] Fix Nova Handler for Consecutive User Prompt Issue

HuanzhiMao:
AWS Bedrock’s Converse API requires an alternating sequence of user and assistant messages. However, certain entries in the live_irrelevance category contain consecutive user prompts, which breaks this requirement. To resolve this, we will apply the `combine_consecutive_user_prompts` process during the `_pre_query_processing_FC` phase to ensure that the chat history alternates properly between user and assistant role.


berkeley-function-call-leaderboard/bfcl/model_handler/api_inference/nova.py
```diff
@@ -74,6 +75,11 @@ def _query_FC(self, inference_data: dict):
         return api_response, end_time - start_time
 
     def _pre_query_processing_FC(self, inference_data: dict, test_entry: dict) -> dict:
+        for round_idx in range(len(test_entry["question"])):
+            test_entry["question"][round_idx] = combine_consecutive_user_prompts(
+                test_entry["question"][round_idx]
```
ShishirPatil, line 80: @HuanzhiMao if the documentation links it we should support this. But for reference, can you also link the documentation in the code as a simple python comment in line 78?

HuanzhiMao: There is no documentation link, but their endpoint will throw this error:
```
An error occurred (ValidationException) when calling the Converse operation: A conversation must alternate between user and assistant roles. Make sure the conversation alternates between user and assistant roles and try again.
```

Fanjia-Yan (approved):
LGTM

PR #887, opened 2025-01-16
add handler for Bielik

dominikabasaj:
Adding Handler for Open Source LLM Bielik available on HF - https://huggingface.co/speakleash/Bielik-11B-v2.3-Instruct .
I chose to load tokenizer in the constructor of the handler, so that I automatically apply chat template. As a result, I can reuse this handler for models with various templates.

### DIFF
--- a/berkeley-function-call-leaderboard/CHANGELOG.md
+++ b/berkeley-function-call-leaderboard/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 All notable changes to the Berkeley Function Calling Leaderboard will be documented in this file.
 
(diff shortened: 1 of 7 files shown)


HuanzhiMao (approved):
Thanks for the PR @dominikabasaj and welcome!

I made a few modifications to your handler.

1. Importing AutoTokenizer from transformers at the top level isn’t ideal because transformers is an optional dependency that some users (e.g., on macOS) may not be able to install. For those who only want to run evaluations for API-based models, this import would break the pipeline.
2. We typically apply the chat template manually so that we can have full control over the final formatted prompt (and that is generally recommended for advanced use cases). 
3. I added a `QuickTestingOSSHandler` so that anyone who prefers to have the chat template auto-applied still has that option.

dominikabasaj:
Thank you!

PR #894, opened 2025-01-27
Falcon3 support

kirill-fedyanin:
Adding FC handler for recently released falcon 3 models with tool calls support (see for example https://huggingface.co/tiiuae/Falcon3-7B-Instruct)
I've run with sglang backend on ast and executable, results look reasonable


berkeley-function-call-leaderboard/bfcl/model_handler/local_inference/falcon_fc.py
```diff
@@ -0,0 +1,94 @@
...
+from bfcl.model_handler.utils import func_doc_language_specific_pre_processing
+from overrides import override
+
+
+class Falcon3FCHandler(OSSHandler):
+    def __init__(self, model_name, temperature) -> None:
+        super().__init__(model_name, temperature)
+        self.model_name_huggingface = model_name.replace("-FC", "")
+        self.bos_token_id = 11
+        self.eos_token_id = 11
```
HuanzhiMao, line 14: What's the purpose of these two? They seem not used.

HuanzhiMao (approved):
Thanks for the PR and welcome @kirill-fedyanin !

PR #895, opened 2025-01-27
Fix minor typo in default system prompt without func

canyon289:
Been working with Huanzhi over email and noticed this typo


berkeley-function-call-leaderboard/bfcl/model_handler/constant.py
```diff
@@ -7,7 +7,7 @@
 If you decide to invoke any of the function(s), you MUST put it in the format of [func_name1(params_name1=params_value1, params_name2=params_value2...), func_name2(params)]
```
canyon289, line 7: On line 4

> If none of the function can be used,

Should this be if none of the **functions** can be used?

HuanzhiMao: Yes, good catch!

HuanzhiMao (approved):
Thanks for the fix and welcome @canyon289 !

PR #897, opened 2025-01-30
[BFCL] Update tool construction for Palmyra models

samjulien:
I've updated `utils.py` in BFCL to reflect some needed changes to how tools are constructed when using Writer's models such as Palmyra X 004.

cc @HuanzhiMao


HuanzhiMao (approved):
Thanks for the fix! LGTM

PR #906, opened 2025-02-10
chore: added snippet for hf datasets compatibility

alt-glitch:


HuanzhiMao (approved):
Thanks for the PR and welcome @alt-glitch !

PR #924, opened 2025-02-28
[BFCL] fix bug in how score_dir is handled for bfcl evaluate

liamcli:
Small fix for when `--score-dir` is passed as an argument to `bfcl evaluate`.


HuanzhiMao (approved):
Thanks for the fix and nice catch!

PR #926, opened 2025-03-01
[BFCL] Support Local Inference for `deepseek-ai/DeepSeek-R1`

HuanzhiMao:
This PR builds on #901 to add support for the `DeepSeek-R1` models by enabling local inference. While #901 introduced inference using the official DeepSeek API endpoint, it is often unstable and overloaded, making it impractical to run the full BFCL benchmarks at the moment (which could take days via API alone). Local inference provides a more efficient alternative for evaluating `DeepSeek-R1` models.

You can use model names to differentiate between the two inference methods:

- `deepseek-ai/DeepSeek-R1` for local inference
- `DeepSeek-R1` for API inference

Note that #901 is a prerequisite for this PR and should be merged first, as it includes the necessary modifications to the base handler that this PR also depends on.


CharlieJCJ (approved):
LGTM

leocnj:
Hi @HuanzhiMao,

I noticed that in the latest BFCL evaluation (08-25, visible on the live leaderboard), only models using `DeepSeekAPIHandler` are included:

```python
"DeepSeek-R1-0528",
"DeepSeek-R1-0528-FC",
"DeepSeek-V3-0324-FC",
```

However, `deepseek-ai/DeepSeek-R1` (which was added via [PR #926](https://github.com/ShishirPatil/gorilla/pull/926)) does not appear to be used in the current evaluation. This model relies on `DeepseekReasoningHandler` and uses a prompting-based approach for generating tool call outputs.

## Questions

### 1. Should we keep `deepseek-ai/DeepSeek-R1`?

Given that the official DeepSeek API now supports native function calling (as used by `DeepSeek-R1-0528-FC`), is there still a use case for maintaining the local inference variant with prompting-based tool calls?

### 2. Recommended handler for locally-hosted models?

If we host a DeepSeek V3 or R1 model locally using SGLang (or similar inference frameworks) that provides an OpenAI-compatible API with **native tool call support**, which handler would you recommend using?

## Context

From PR #926, I understand the original motivation was API instability and long benchmark times. However, if we're hosting the model ourselves with an OpenAI-compatible endpoint that supports native function calling, it seems like we could potentially use a different handler that takes advantage of native tool call features rather than the prompting-based approach.

Thank you for clarifying!

PR #933, opened 2025-03-08
Move logic to eval a task in a separate function.

fvisin:
This change exposes an API for 3P code to run the evaluation of a single task, while also improving code readability.


HuanzhiMao (approved):
Tested and works as expected. LGTM. Thank you for the PR @fvisin !

PR #937, opened 2025-03-13
Fix Gorilla Paper `requirements.txt` Location to Remove Global Dependency Confusion

HuanzhiMao:
This PR addresses a legacy issue where the original requirements.txt for the Gorilla paper (introduced in #61) was not placed inside the gorilla folder. Previously, this caused confusion, as it appeared to represent global dependencies. This fix moves the file into the correct location to eliminate ambiguity.


CharlieJCJ (approved):
Good catch, LGTM

PR #938, opened 2025-03-13
[BFCL] Add _unused Suffix to Unused Dataset Files in the BFCL Benchmark

HuanzhiMao:
This PR updates dataset filenames for the categories not currently included in the BFCL benchmark (sql, chatable, and multi_turn_composite). By appending an _unused suffix to each of these unused dataset files, we clarify that they are excluded from the current benchmark iteration and prevent potential confusion going forward.


CharlieJCJ (approved):
LGTM, thanks

PR #939, opened 2025-03-13
[BFCL] Add Google `Gemma-3` Series Models

HuanzhiMao:
  - `google/gemma-3-1b-it`
  - `google/gemma-3-4b-it`
  - `google/gemma-3-12b-it`
  - `google/gemma-3-27b-it`


HuanzhiMao:
cc @canyon289 @fvisin

HuanzhiMao:
cc @canyon289 @fvisin 
sglang/vllm are still in the process of adding support for the Gemma-3 models, which should be completed in the next few days. We won’t be able to run evaluations until the serving frameworks fully support them.

canyon289:
Thanks for that update

canyon289:
When you run these btw I would be interested in the full response and score files. Like discussed before no need to add them to the repo because of size, but if you have them on a one time copy would be awesome!

PR #952, opened 2025-03-20
Use tree-sitter-languages to reduce dependencies.

fvisin:
Tree-sitter-languages conveniently exposes all languages at once, removing the need to depend on language specific tree-sitter extensions.


HuanzhiMao (commented):
Thanks for the PR @fvisin !
I noticed that the `tree_sitter_language` package ([Link](https://github.com/grantjenks/py-tree-sitter-languages?tab=readme-ov-file)) is marked as unmaintained:
> This library is unmaintained, as an alternative you can use .. tree-sitter-language-pack: https://github.com/Goldziher/tree-sitter-language-pack

Should we consider switching to `tree-sitter-language-pack` instead?

fvisin:
Ah, bummer! Sorry I didn't realise that it was unmaintained. Then yes, absolutely, let's use the updated fork instead!

PR #959, opened 2025-03-25
[BFCL] Fix API Keys Handling

catherineruoxiwu:
The following bugs can all be reproduced from a clean, unmodified version of the repository. All of them have been addressed by this PR.
#### Setup 1:  
- **Input:** `test_case_ids_to_generate.json` → `rest: ["rest_0"]`  
- **.env:** No API key provided  
- **Command:** `bfcl generate --model gpt-4o-mini-2024-07-18 --run-ids`

1. fix `check_api_key_supplied` implementation in `utils.py`. `os.getenv(var)` returns `None` instead of a string

2. After the previous issue is fixed, a `NameError` occurs when no result file existed and `existing_ids` was only conditionally defined inside a loop
   ```
   Exception has occurred: NameError  
   free variable 'existing_ids' referenced before assignment in enclosing scope
   ```
3. Under this setup, no test should be called, but test case `rest_0` was still collected unexpectedly. 
Screenshot 2025-03-25 at 2 17 05 AM

#### Setup 2:  
- **.env:** All API keys provided  
- **Command:** `bfcl generate --model gpt-4o-mini-2024-07-18 --test-category rest`

4. API keys were replaced in the rest tests JSON file, but the credential information in the question passed to the `generate_result` function was not correct.

### DIFF
--- a/berkeley-function-call-leaderboard/bfcl/utils.py
+++ b/berkeley-function-call-leaderboard/bfcl/utils.py
@@ -198,7 +198,7 @@ def check_api_key_supplied() -> bool:
     """
     ENV_VARS = ("GEOCODE_API_KEY", "RAPID_API_KEY", "OMDB_API_KEY", "EXCHANGERATE_API_KEY")
     for var in ENV_VARS:
-        if os.getenv(var) == "":
+        if os.getenv(var) is None:
             return False
     return True
 
--- a/berkeley-function-call-leaderboard/bfcl/_llm_response_generation.py
+++ b/berkeley-function-call-leaderboard/bfcl/_llm_response_generation.py
@@ -72,6 +72,8 @@ def get_involved_test_entries(test_category_args, run_ids):
     all_test_file_paths, all_test_categories, all_test_entries_involved = [], [], []
     api_key_supplied = check_api_key_supplied()
     skipped_categories = []
+    rest_test_credential_applied = False
+
     if run_ids:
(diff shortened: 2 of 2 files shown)


berkeley-function-call-leaderboard/bfcl/utils.py
```diff
@@ -198,7 +198,7 @@ def check_api_key_supplied() -> bool:
     """
     ENV_VARS = ("GEOCODE_API_KEY", "RAPID_API_KEY", "OMDB_API_KEY", "EXCHANGERATE_API_KEY")
     for var in ENV_VARS:
-        if os.getenv(var) == "":
+        if os.getenv(var) is None:
```
HuanzhiMao, line 201: 1. When the user copies the `.env.example` content to `.env` without filling in the actual key (eg, `RAPID_API_KEY=`), `os.getenv("RAPID_API_KEY")` will return `""` (empty string). 
2. When the user has no `.env` file or there is no entry on the `RAPID_API_KEY` key, `os.getenv("RAPID_API_KEY")` will return `None`. 

Since both scenarios are possible, you should handle both situations.

HuanzhiMao (changes requested):

berkeley-function-call-leaderboard/bfcl/utils.py
catherineruoxiwu: Thank you for pointing that out! Just fixed it

HuanzhiMao (approved):
LGTM

PR #960, opened 2025-03-25
[BFCL] Use HTTPS instead of HTTP for OMDB

hrshtv:
Executable evals fail sometimes with the following returned as the response: `{"Response":"False","Error":"Invalid API key!"}`
See this for context: https://github.com/omdbapi/OMDb-API/issues/311#issuecomment-2138657319.


HuanzhiMao (approved):
Thank you for the PR and welcome @hrshtv !

PR #962, opened 2025-03-26
[BFCL] Fix ambiguity in exec_parallel_10 question

amitojsingh2022:
Fixed the dataset issue by changing the "feet" and "inches" to meters for the second and third triangle area question in exec_parallel_10.


HuanzhiMao (approved):
LGTM

PR #966, opened 2025-03-28
[BFCL] Multi-turn TravelAPI book_flight() Fix

amitojsingh2022:
This PR addresses the issue raised here: https://github.com/ShishirPatil/gorilla/discussions/882
To create the fix, we remove the travel_cost parameter from book_flight() and add validation and cost calculation logic within the function to address the concern of ambiguity raised in the issue. The issue was that there were two different interpretations of the parameter: 1. the travel cost should be calculated, 2. using an arbitrary user-specified value. 

The updates in the implementation calculate the travel cost by calling get_flight_cost() internally, so it eliminate the confusion and sticks with the first interpretation. In addition, I updated [multi_turn_base_169] in order to match with this change and clear any ambiguity.


HuanzhiMao (changes requested):
Thanks for the PR. 
Since you changed the function signature for `book_flight`, you need to check and update all multi-turn entries that used the `book_flight` function. The ground truth needs to be updated so that they are not in invalid syntax, and the user query might also need to be refined/reworded to make it sound more natural.

catherineruoxiwu:
LGTM!

HuanzhiMao (approved):

PR #1032, opened 2025-05-24
Add support for llama-3.1-nemotron-ultra-253b-v1 to BFCL

AdityaGhai18:
Add LLaMA-3.1 Nemotron Ultra 253B v1 (FC) to the leaderboard.

Used Nvidia API inference handler


berkeley-function-call-leaderboard/bfcl/constants/model_config.py
```diff
@@ -718,6 +718,18 @@ class ModelConfig:
...
+    "nvidia/llama-3.1-nemotron-ultra-253b-v1": ModelConfig(
+    model_name="nvidia/llama-3.1-nemotron-ultra-253b-v1",
+    display_name="LLaMA-3.1 Nemotron Ultra 253B v1 (FC)",
+    url="https://build.nvidia.com/nvidia/llama-3_1-nemotron-ultra-253b-v1/modelcard",
+    org="NVIDIA",
+    license="Proprietary",
+    model_handler=NvidiaHandler,
+    input_price=None,
+    output_price=None,
+    is_fc_model=True,
```
ShishirPatil, line 730: The "BFCLv2" part of the [modelcard](https://build.nvidia.com/nvidia/llama-3_1-nemotron-ultra-253b-v1/modelcard) seems to indicate it is prompting mode? So, shouldn't `if_fc_model` be `false`?

HuanzhiMao: It's not exactly the same as the normal bfcl prompting method. From their model card, they used the following system prompt during evaluation. I bolded the part that are different; note that we don't have such **\<TOOLCALL>** or **<AVAILABLE_TOOLS>** in our [default system prompt](https://github.com/ShishirPatil/gorilla/blob/main/berkeley-function-call-leaderboard/bfcl/constants/default_prompts.py#L3-L18). 

----

You are an expert in composing functions. You are given a question and a set of possible functions.
Based on the question, you will need to make one or more function/tool calls to achieve the purpose.
If none of the function can be used, point it out. If the given question lacks the parameters required by the function,
also point it out. You should only return the function call in tools call sections.

If you decide to invoke any of the function(s), you MUST put it in the format of **\<TOOLCALL>**[func_name1(params_name1=params_value1, params_name2=params_value2...), func_name2(params)]**\</TOOLCALL>**

You SHOULD NOT include any other text in the response.
Here is a list of functions in JSON format that you can invoke.

**<AVAILABLE_TOOLS>**{functions}**</AVAILABLE_TOOLS>**

{user_prompt}

HuanzhiMao (changes requested):
It's not exactly the same as the normal bfcl prompting method. From their model card, they used the following system prompt during evaluation. I bolded the part that are different; note that we don't have such **\<TOOLCALL>** or **<AVAILABLE_TOOLS>** in our [default system prompt](https://github.com/ShishirPatil/gorilla/blob/main/berkeley-function-call-leaderboard/bfcl/constants/default_prompts.py#L3-L18). 

----

You are an expert in composing functions. You are given a question and a set of possible functions.
Based on the question, you will need to make one or more function/tool calls to achieve the purpose.
If none of the function can be used, point it out. If the given question lacks the parameters required by the function,
also point it out. You should only return the function call in tools call sections.

If you decide to invoke any of the function(s), you MUST put it in the format of **\<TOOLCALL>**[func_name1(params_name1=params_value1, params_name2=params_value2...), func_name2(params)]**\</TOOLCALL>**

You SHOULD NOT include any other text in the response.
Here is a list of functions in JSON format that you can invoke.

**<AVAILABLE_TOOLS>**{functions}**</AVAILABLE_TOOLS>**

{user_prompt}

AdityaGhai18:
Thank you for highlighting the prompt format requirements.

I've added a custom NemotronHandler to support the required XML-based function calling format (<TOOLCALL> and <AVAILABLE_TOOLS> tags) and updated the model config accordingly.
The system prompt below is taken directly from the model card, as you pointed out.

I also implemented the necessary handler methods (_pre_query_processing_prompting, add_first_turn_message_prompting, decode_ast, and decode_execute) to ensure the prompt is formatted, sent, and parsed correctly according to the model’s requirements.

If there’s anything else I might have missed or if you have further changes, please let me know. Thanks again for your help!

PR #1033, opened 2025-05-25
[BFCL] Remove latency data for open source models

errorfourten:
This PR removes latency statistics for models that do not have a input/output cost (typically open source models). This is because latency for these open source models usually depend on the hardware that they are run on.


HuanzhiMao (approved):
Thanks for the PR @errorfourten and welcome!

PR #1038, opened 2025-05-27
[BFCL] Add `version` to `bfcl` CLI

ShishirPatil:
Currently bfcl CLI supports `models`, `test-categories`, `generate`, `results`, `evaluate`, `scores`. Adding `version`. 

Once we set up distribution (also) via PyPI, the version command will align with the PyPI package version. However, for now, we’re still using the commit hash to establish provenance, as shown on the BFCL website.
I believe continuing to rely on the commit hash remains the best approach for now. The addition of versioning will primarily help with PyPI integration, making it easier to pull updates locally when new versions are published.

image


HuanzhiMao (approved):
💯

PR #1048, opened 2025-06-04
Add Support for Using Gemini API to Access Google Models

Guangyu-Joshua-Feng:
- Dual backend support: Added compatibility for both the Vertex AI SDK and the Google GenAI (Gemini) SDK, with backend selection configurable by .env.
- Json Function doc Schema fix: Replaced unsupported "format": "date" with "format": "date-time" in function JSON schemas to comply with the GenAI schema specification.
- Function doc format caution for GenAI: The GenAI SDK enforces stricter validation: all fields listed in "required" must be explicitly defined under "properties", including nested objects.

> Thus, the following test case fails under GenAI due to missing required subfields in a nested object schema:
> File: BFCL_v3_parallel.json
> Test Case ID: parallel_29
> Issue: population lists "adults", "children", "singles" as required, but these fields were not defined in its properties block.


HuanzhiMao (changes requested):
Thanks for the PR @Guangyu-Joshua-Feng !

Given that it's completely different logic for using AI Studio vs. Vertex AI, I think it would be better to put them in two different handlers. Eg, something like `gemini_vertex_ai.py` and `gemini_ai_studio.py`. By default, the model config use the AI Studio handler, but user can easily update the config to switch to Vertex AI.

Regarding `parallel_29`, we should transform that function doc to valid format in our dataset.

Guangyu-Joshua-Feng:
GeminiHandler code is refactored into two subclasses and one base class.

PR #1050, opened 2025-06-05
fix: loosen openai requirements to be >= 1.76.0

TheFloatingString:
Adding a more flexible dependency requirement for the `openai` package for the Berkeley Function Calling Leaderboard v3 (BFCL-v3), making it easier to use `bfcl` as a python package in software projects that require higher `openai` versions.


HuanzhiMao (approved):
Valid point. Thanks for the PR!

PR #1057, opened 2025-06-09
[BFCL] Replace Exception with SyntaxError for Java and JavaScript Parsers

TheFloatingString:
for the Java and JavaScript parsers, I updated the error handling such that it returns a `SyntaxError` rather than an `Exception`.

### Motivations
- this is done in parallel with using `bfcl` to evaluate models, raising a specific type of error works better to handle erroneous model responses (as opposed to runtime errors that are unrelated to the model itself)

### Limitations
- `SyntaxError` may be too narrow of a scope

if there is another error type that is more relevant, I'm glad to make the changes needed on my side

thanks in advance!


HuanzhiMao (approved):
Thanks for the PR @TheFloatingString !

PR #1058, opened 2025-06-10
[BFCL] Add type hinting

TheFloatingString:
adding function param type hinting for the decoding functions so it's a bit easier to run function calls in a stand-alone way,

hope this helps!:)


HuanzhiMao (approved):
Thanks for adding the type hints @TheFloatingString !

PR #1061, opened 2025-06-11
[BFCL] Support DashScope API Inference for `Qwen3` Series

HuanzhiMao:
This PR allows users to run the `Qwen3` series through Alibaba's official hosted API (dashscope.aliyuncs.com), in addition to the self-hosted option. 
The hosted API can dynamically extend the context window for longer prompts and can take reasoning content for chat history, which is nice.


CharlieJCJ (approved):
LGTM

PR #1062, opened 2025-06-12
[BFCL] Introduce OpenAI Responses API Handler + o4-mini/o3 models

errorfourten:
This PR adds a new handler for the Responses API that OpenAI [introduces](https://platform.openai.com/docs/guides/responses-vs-chat-completions).

Main changes:
* Added a new `OpenAIResponsesHandler`. This is done as many models still rely on the Completions API and we do not want to break comptability.
* Updating of inaccurate type hinting. Changed from `any` type hints to `typing.Any`
* Modified function calling and reasoning output to be included in the conversation context
* Removed o1 and o3-mini models
* Introduced o4-mini and o3 models


berkeley-function-call-leaderboard/bfcl_eval/model_handler/model_style.py
```diff
@@ -5,6 +5,7 @@
 class ModelStyle(Enum):
     Gorilla = "gorilla"
     OpenAI = "gpt"
```
errorfourten, line 7: Please let me know if it would be clearer to rename this to `OpenAICompletions` or if it's okay as-is

HuanzhiMao: Yea let's use `OpenAI_Completions` and `OpenAI_Responses`

HuanzhiMao (approved):
Thanks for the PR @errorfourten !

PR #1067, opened 2025-06-17
Documentation added for model handler

Aktsvigun:
Documentation added for the `model_handler` part of the BFCL benchmark

### DIFF
--- a/berkeley-function-call-leaderboard/bfcl_eval/model_handler/api_inference/claude.py
+++ b/berkeley-function-call-leaderboard/bfcl_eval/model_handler/api_inference/claude.py
@@ -23,12 +23,25 @@ from bfcl_eval.utils import is_multi_turn
 
 
 class ClaudeHandler(BaseHandler):
-    def __init__(self, model_name, temperature) -> None:
+    """
+    A handler class for interacting with Anthropic's Claude models, supporting both function calling and standard prompting approaches. This class inherits from BaseHandler and implements model-specific processing for Claude API interactions.
+    """
+    def __init__(self, model_name: str, temperature: float) -> None:
         super().__init__(model_name, temperature)
         self.model_style = ModelStyle.Anthropic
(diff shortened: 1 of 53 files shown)


berkeley-function-call-leaderboard/bfcl_eval/model_handler/api_inference/claude.py
```diff
@@ -23,12 +23,25 @@
...
+    """
+    A handler class for interacting with Anthropic's Claude models, supporting both function calling and standard prompting approaches. This class inherits from BaseHandler and implements model-specific processing for Claude API interactions.
+    """
+    def __init__(self, model_name: str, temperature: float) -> None:
         super().__init__(model_name, temperature)
         self.model_style = ModelStyle.Anthropic
         self.client = Anthropic(api_key=os.getenv("ANTHROPIC_API_KEY"))
 
-    def decode_ast(self, result, language="Python"):
+    def decode_ast(self, result: str, language: str="Python") -> list[dict[str, dict]]:
```
HuanzhiMao, line 34: `result` is not guaranteed to be a string. When in prompting mode, it is indeed a string. However, when in FC mode, it would be a list of dictionaries; you can infer from lines 58-60 (cited below).
```
for invoked_function in result:
    name = list(invoked_function.keys())[0]
    params = json.loads(invoked_function[name])
```

HuanzhiMao:
Hey @Aktsvigun ,

One thing I'm thinking: Because each concrete handler exposes the same methods with identical parameters and behavior, it might be cleaner to keep the detailed documentation in BaseHandler rather than duplicating it everywhere. IntelliSense should be able to surface the parent docstring when we hover over the overridden methods, so developers won’t lose any context. For any extra helper functions unique to a given handler, we can certainly document those in place.

What do you think? Does that approach sound reasonable to you?

Aktsvigun:
Agree, I didn't know they share the same exact methods! I'll relaunch, thank you for your comments here.

PR #1068, opened 2025-06-18
[BFCL] Fix prompt concatenation bug in Qwen template

nehcgs:
This PR fixes an issue where the special token `<|im_end|>\n` was being incorrectly added using. This operation does not modify `formatted_prompt` in-place, which can lead to the token not being appended as expected. 

**Changes**
Replaced `formatted_prompt + "<|im_end|>\n"` with `formatted_prompt += "<|im_end|>\n"` to ensure proper in-place string concatenation.

**Note**
This issue significantly affects `Qwen3` model performance in the leaderboard.

### DIFF
--- a/berkeley-function-call-leaderboard/CHANGELOG.md
+++ b/berkeley-function-call-leaderboard/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 All notable changes to the Berkeley Function Calling Leaderboard will be documented in this file.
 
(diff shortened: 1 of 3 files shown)


HuanzhiMao (approved):
Thanks for the fix @nehcgs ! Nice catch :)

Just a heads-up: the leaderboard’s Qwen3 score has its inference done through Alibaba’s hosted API, so merging this PR **will not** affect the score currently on the leaderboard; it only changes the score for the self-hosted version.

PR #1071, opened 2025-06-20
feat: type hinting for runner functions

TheFloatingString:
adding some type hinting for the runner functions


berkeley-function-call-leaderboard/bfcl_eval/eval_checker/eval_runner.py
```diff
@@ -33,8 +34,14 @@ def get_handler(model_name):
 
 
 def multi_turn_runner(
-    handler, model_result, prompt, possible_answer, model_name, test_category, score_dir
-):
+    handler: str,
```
HuanzhiMao, line 37: handler should be of type `BaseHandler` (from `berkeley-function-call-leaderboard/bfcl_eval/model_handler/base_handler.py`)

PR #1072, opened 2025-06-21
Add Qwen handler

zhangyingerjelly:
Added QwenAgentThinkHandler and QwenAgentNoThinkHandler in the file api_inference/qwen.py


HuanzhiMao (approved):
Thanks for the PR @zhangyingerjelly !

noiji:
@zhangyingerjelly hi!

I was just wondering why you have added agent handlers where vllm needs to be run while oss model is already supported in [qwen_fc.py](https://github.com/ShishirPatil/gorilla/blob/main/berkeley-function-call-leaderboard/bfcl_eval/model_handler/local_inference/qwen_fc.py#L12) :)

(I am trying to evaluate local models and was weighing on which handlers to use)

PR #1078, opened 2025-06-25
[BFCL] Add support for Arch-Agent

nehcgs:
This PR adds Arch-Agent-{1.5B, 3B, 7B, 32B} to the leaderboard. 

**Note**
Note that we use YaRN scaling to deploy the models, and all Arch-Agent models are evaluated with a context length of 64K. In the current codebase, this requires adding extra fields for YaRN scaling when spinning up servers and manually overwrite `self.max_context_length` in the `base_oss_handler.py`.

### DIFF
--- a/berkeley-function-call-leaderboard/bfcl_eval/constants/model_config.py
+++ b/berkeley-function-call-leaderboard/bfcl_eval/constants/model_config.py
@@ -47,6 +47,7 @@ from bfcl_eval.model_handler.local_inference.qwen_fc import QwenFCHandler
 from bfcl_eval.model_handler.local_inference.salesforce_llama import SalesforceLlamaHandler
 from bfcl_eval.model_handler.local_inference.salesforce_qwen import SalesforceQwenHandler
 from bfcl_eval.model_handler.local_inference.think_agent import ThinkAgentHandler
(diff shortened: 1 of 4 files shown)


berkeley-function-call-leaderboard/bfcl_eval/constants/model_config.py
```diff
@@ -1954,6 +1955,54 @@ class ModelConfig:
...
+        output_price=None,
+        is_fc_model=True,
+        underscore_to_dot=False,
+    ),
+    "katanemo/Arch-Agent-32B": ModelConfig(
+        model_name="katanemo/Arch-Agent-32B",
+        display_name="Arch-Agent-32B",
+        url="https://huggingface.co/katanemo/Arch-Agent-32B",
+        org="katanemo",
+        license="Proprietary",
```
HuanzhiMao, line 1999: According to the [HuggingFace model card](https://huggingface.co/katanemo/Arch-Agent-32B), the license should be `katanemo-research`?

nehcgs: Updated, thanks for reviewing! Please let me know if you have any other concerns.

HuanzhiMao (approved):
Thanks for the PR @nehcgs !

PR #1085, opened 2025-06-29
Update irrelevance_232 question 

Gnav3852:
Change the question in irrelevance_232 from: "Find a hotel in New York that provides breakfast and has a fitness centre." to: "Find a yoga retreat in New York that includes vegetarian meals and spa services."


HuanzhiMao (approved):
Thanks for the fix @Gnav3852 !

PR #1086, opened 2025-06-29
[BFCL] Resolve duplicated 'live-relevance_3-3-0' test entry id

eungizoang:
There are two test_entries with the same ID: `live-relevance_3-3-0`.
Although this doesn't cause any errors during evaluation, I found that the duplication could lead to confusion later on.
Therefore, I changed the ID of the second entry to `live-relevance_17-17-0`.


HuanzhiMao (approved):
Good catch. Thanks for the fix and welcome @eungizoang !

PR #1087, opened 2025-06-29
[BFCL] Fixed missing airport route entries

amitojsingh2022:
Added 3 new airport routes for base costs in travel_booking.py:

```
("LAX", "HND"): 430,
("CRH", "HKG"): 290,
("CRH", "JFK"): 300
```


HuanzhiMao (approved):
Thanks for the fix!

PR #1088, opened 2025-06-30
Add DM-Cito-8B-v2

kevin2016:
This PR adds model handler for DM-Cito-8B-v2. The url of the model api is：https://rapidapi.com/kechengkevin/api/dm-cito-8b-v2

Meanwhile remove the model  DM-Cito-8B

Note:
Because our server has limited performance, please don't set the number of threads too high. Using --num-threads 30 worked without any issues during our tests.


HuanzhiMao (approved):
Thanks for the PR!

PR #1090, opened 2025-07-01
[BFCL] Replaced system role with developer role

errorfourten:
OpenAI's [latest model spec](https://model-spec.openai.com/2025-04-11.html#definitions) introduces the `developer` role which is preferred over using the `system` role.

However, it is probably not safe to make direct changes to the datasets prompts due to other models potentially not supporting this new role. As such, a new function was introduced to convert the roles from `system` into `developer` only for OpenAI models


HuanzhiMao (approved):
Thanks for the PR @errorfourten !

PR #1098, opened 2025-07-06
[BFCL] Reintroduce latency stats for local models, update cost calculation

### LINKED ISSUE #1035: [BFCL] Re-introduce latency data for open source models
**Describe the issue**
PR #1033 got rid of latency for open-source models. However, I think there is value in having latency for open-source models. This is because: 
1. The relative ordering is useful, and relevant since we run all models on the same hardware fleet. 
2. The next obvious question is weather we can use proxy such as number tokens to measure latency -  with reasoning models, this is not a good proxy. 

**Proposed Changes**

Revert #1033 ? 

Although we should discuss if there exists a better solution.

Gnav3852:
Resolves #1035 

  - Re-introduce latency statistics for locally hosted models
  - Update cost calculation to cover the entire dataset batch, instead of the average cost per 1k function calls
  - For local-hosted models, we compute total cost by summing all latencies and multiplying by the hourly GPU price.


ShishirPatil:
Thank you @Gnav3852 !

Issue #914, opened 2025-02-20
[BFCL] Multi-turn evaluation problem.


lucenzhong:
For `multi_turn_base_186` at turn 3, the user's question is

> There has been a problem with my booking and I previously reached out to support without any feedback yet. Kindly contact customer support on my behalf, emphasizing the smooth facilitation of my travel arrangements.

The model is supposed to call the `customer_support_message` function with a specific `user message`. I didn't find any definition of the standard message definition, as I understand it, the meesage parameter is a generated by the model.

"contact_customer_support(booking_id='3426812',message='I am experiencing an issue with my booking and require immediate assistance to ensure the smooth facilitation of my travel arrangements. Please address this matter urgently.')"

lucenzhong:
For `multi_turn_base_79` at turn 2. 
My result:
```
"setCruiseControl(activate=True,distanceToNextVehicle=100.0,speed=65.0)"
```
Ground truth:
```
"setCruiseControl(speed=65, activate=True, distanceToNextVehicle=100)"`.
```
error message:
```
Model response execution results so far does not contain all the ground truth response execution results for turn 2."
```
Since the `setCruiseControl` function define `speed` and `distanceToNextVehicle` as float.  I think the gound truth may be wrong.
```
  def setCruiseControl(
      self, speed: float, activate: bool, distanceToNextVehicle: float
  ) -> Dict[str, Union[str, float]]:
      """
      Sets the cruise control of the vehicle.
      Args:
          speed (float): The speed to set in m/h. The speed should be between 0 and 120 and a multiple of 5.
          activate (bool): True to activate the cruise control, False to deactivate.
          distanceToNextVehicle (float): The distance to the next vehicle in meters.
      Returns:
          cruiseStatus (str): The status of the cruise control. [Enum]: ["active", "inactive"]
          currentSpeed (float): The current speed of the vehicle in km/h.
          distanceToNextVehicle (float): The distance to the next vehicle in meters.
      """
```

lucenzhong:
For `multi_turn_base_24` at turn 0, the user's question is:
> Look for draft and final report in my current directory. Compare the content difference of both.

The ground_truth is 
```
diff(file_name1='report_draft.txt', file_name2='report_final.txt')
```

The `report_draft.txt` and `report_final.txt` is not specified in the question. I don't think it's reasonable to let the model to guess the full file name.

HuanzhiMao:
Hey @lucenzhong,

[...]

PR #1110, opened 2025-07-14
[BFCL] Contact Customer Support Multi Turn & Vehicle Control #914

amitojsingh2022:
Fixed Issue 914. 
1. Removed message from the Contact Customer Support Function and updated entry questions to remove message specificity prompt.
2. Cast everything to float before returning in setCruiseControl function in Vehicle Travel. Solved ambiguity with int/float conversion.


HuanzhiMao (approved):
Thanks for the fix!

PR #1126, opened 2025-07-21
Fix qwen3 inference data and qwenfc handler chat template

Shanay-Mehta:
1. Reasoning content was not being added in the inference data for both QwenHandler and QwenFCHandler.
2. The _format_prompt function in QwenFCHandler was not checking for tool_calls in assistant messages. All those messages where the assistant requested a tool call were converted to empty messages.

This PR fixes both the above points.

This also means that all those scores for qwen3-prompt and qwen3-FC that were calculated using local_inference, would need to be calculated again.


HuanzhiMao (approved):
Thanks for the fix and nice catch @Shanay-Mehta !

PR #1130, opened 2025-07-24
updated the model_config for qwen, grok, mistral model names

zimo0110:
New names for qwen models, updated names for grok (beta is gone) added the newest model of grok4, mistral names updated too


HuanzhiMao (approved):
Thanks for the PR @zimo0110 !

PR #1131, opened 2025-07-24
added api support for moonshotai kimi-k2

zimo0110:
Added API support only for kimi k2 model, added a handler class for it and added it to the model_config


HuanzhiMao (approved):
Thanks for the PR @zimo0110 !

PR #1019, opened 2025-05-11
BFCL V4 Release

HuanzhiMao:
> ❗️**Important**: This PR introduces breaking changes and is **NOT** backward-compatible.

# BFCL V4

💥 At ICML 2025, we’re delighted to introduce BFCL V4 Agentic, a new benchmark focused on tool-calling in real-world agentic settings — including:

🔍 Web search with multi-hop reasoning and error recovery
🧠 Evaluating Tool-Calling for Memory
⚠️ Evaluating Format Sensitivity 

## Change Log

1. **New agentic domain**
   - Introduces the agentic domain with two categories: Web Search and Memory Management.
   - For more information, please see our accompanying [blog posts](https://gorilla.cs.berkeley.edu/blog.html).
2. **Revised overall-accuracy formula**

   - As single-turn tasks approach saturation, weighting now favors complex, multi-step agentic tasks.

   | Segment     | Old % |  New % |
   | ----------- | ----: | -----: |
   | Live        |    33 | **10** |
   | Non-Live    |    33 | **10** |
   | Irrelevance |     0 | **10** |
   | Multi-Turn  |    33 | **30** |
   | Agentic     |     0 | **40** |

3. **Leaderboard / model cleanup**
   - Retires several deprecated models from the leaderboard.
   - Removes unused model handlers to improve maintainability.
4. **Address #602**
   - `Non-Live Acc` and `Live Acc` score calculation now excludes the Irrelevance/Relevance category scores.
5. **Resolve #1094.**
6. **Codebase refactor**
   - Reorganizes the response-generation pipeline and related modules for easier maintenance.
   - Simplify the response-generation pipeline logic for locally-hosted models.
   - Introduce `enums.py`
7. **Test category rename**
   The following categories have been renamed to avoid confusion. This applies to both dataset file names and leaderboard website columns.
   - `simple` --> `simple_python`
   - `java` --> `simple_java`
   - `javascript` --> `simple_javascript`
8. **Directory layout overhaul**
   Results and scores now use a _two-level_ hierarchy:

   ```text
   result/<model>/<general_category>/<category>.json
   score/<model>/<general_category>/<category>.json
   ```

   `general_category` ∈ { **non_live**, **live**, **multi_turn**, **agentic**, **format_sensitivity** }

   • For _agentic-memory_ tasks, an extra level distinguishes the memory backend:

   ```text
   result/<model>/agentic/<memory_backend>/<category>.json
   ```

[...]

PR #1150, opened 2025-08-16
Fix bfcl-generate: Error for OAI non-reasoning models

shivamMg:
Running `bfcl generate` with OpenAI non-reasoning models ex. gpt-4.1-mini, results in requests failing with: `Encrypted content is not supported with this model.`.

Fix it by adding reasoning related parameters only in case of reasoning models.

### DIFF
--- a/berkeley-function-call-leaderboard/bfcl_eval/model_handler/api_inference/openai_response.py
+++ b/berkeley-function-call-leaderboard/bfcl_eval/model_handler/api_inference/openai_response.py
@@ -79,11 +79,12 @@ class OpenAIResponsesHandler(BaseHandler):
         }
 
         if self._is_reasoning_model():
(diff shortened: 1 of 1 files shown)


shivamMg:
@ShishirPatil would love to get a review

ShishirPatil:
Hey @shivamMg just came across this. 
Yeah agree. Do you mind having a generic set-up. Now to also include o5-* ?

ShishirPatil:
Similar to #1187

HuanzhiMao:
Thanks for the contribution, @shivamMg! We’ve already addressed this in #1019 as part of the BFCL v4 release, which removes the reasoning-only params for non-reasoning OpenAI models. Since the fix is in v4, we’re going to close this PR.

If you still see issues on BFCL v4 or later, please let us know and we can reopen. 🙏

PR #1160, opened 2025-08-31
[BFCL] 

amitojsingh2022:


JakobGr:
Thanks @amitojsingh2022 and @HuanzhiMao ! The only thing I do not agree with is the implementation of the echo function.  You are right that according to the function documentation, the echo command is said not to create new files — implying that touch must be called beforehand.
However, the current implementation appears to contradict this:

```
if file_name in self._current_dir.contents:
    self._current_dir._get_item(file_name)._write(content)
else:
    self._current_dir._add_file(file_name, content)
```

If the file doesn’t exist, echo directly calls _add_file(file_name, content), effectively creating a new file with the provided content.

Proposed fix:
To align the implementation with the intended behavior (that echo should only write to existing files and require touch first), the relevant section could be modified to return an error when the file does not exist — for example:

```
if file_name in self._current_dir.contents:
    self._current_dir._get_item(file_name)._write(content)
else:
    return {"error": f"echo: cannot write to '{file_name}': No such file"}
```

HuanzhiMao:
> Thanks @amitojsingh2022 and @HuanzhiMao ! The only thing I do not agree with is the implementation of the echo function. You are right that according to the function documentation, the echo command is said not to create new files — implying that touch must be called beforehand. However, the current implementation appears to contradict this:
> 
> ```
> if file_name in self._current_dir.contents:
>     self._current_dir._get_item(file_name)._write(content)
> else:
>     self._current_dir._add_file(file_name, content)
> ```
> 
> If the file doesn’t exist, echo directly calls _add_file(file_name, content), effectively creating a new file with the provided content.
> 
> Proposed fix: To align the implementation with the intended behavior (that echo should only write to existing files and require touch first), the relevant section could be modified to return an error when the file does not exist — for example:
> 
> ```
> if file_name in self._current_dir.contents:
>     self._current_dir._get_item(file_name)._write(content)
> else:
>     return {"error": f"echo: cannot write to '{file_name}': No such file"}
> ```

Thanks for pointing this out! I agree with your reasoning here. This has been addressed in #1206 .

PR #1170, opened 2025-09-09
[Not For Checkin] llguidance experimentation

riedgar-ms:
Work on getting llguidance working with BFCL


bbrowning:
@riedgar-ms For other reasons I recently had a grammar that worked fairly well in local testing for doing Qwen model style tool calling with json schemas - see gist at https://gist.github.com/bbrowning/428b6a26f9f419ae672f38621eaf819c for an example of how I converted the `tools` parameter from a ChatCompletion request into a guided grammar for use with Hermes style JSON tool calls, which looks like what you were attempting here.

PR #1173, opened 2025-09-10
Adjust `num_threads` handling in `bfcl generate`

riedgar-ms:
Tweak the `--num-threads` argument to `bfcl generate`, so that if it is not set then `args.num_threads` will be None, and handle this explicitly in `generate_results()`


riedgar-ms:
When trying to debug, I tend to request fewer threads. It is confusing when I request 1 thread and get 100.

PR #1177, opened 2025-09-13
Fix ground truth for multi_turn_base_154: merge turns 2 and 3

chughtapan:
## Summary
This PR fixes a ground truth data inconsistency in test case `multi_turn_base_154` that was causing an IndexError during evaluation.

## The Problem
- Test has **5 user questions** (5 turns of interaction)
- Ground truth incorrectly expected **6 turns** of function calls
- The evaluator fails with `IndexError: list index out of range` when trying to access turn 6

## Root Cause
The second user question asks to both find the airport AND check airfare:
> "Let's identify the nearest airport to Chicago first, **then** we'll continue to check airfare from there to Los Angeles for Aug 10th 2024."

The ground truth incorrectly split this single request into two separate turns:
- Turn 2: `get_nearest_airport_by_city(location='Chicago')`
- Turn 3: `get_flight_cost(...)`

## The Fix
Merged ground truth turns 2 and 3 into a single turn to match the actual user interaction pattern:

**Before (6 turns for 5 questions):**
```python
Turn 1: verify_traveler_information
Turn 2: get_nearest_airport_by_city
Turn 3: get_flight_cost  
Turn 4: set_budget_limit
Turn 5: book_flight
Turn 6: view_messages_sent
```

**After (5 turns for 5 questions):**
```python
Turn 1: verify_traveler_information
Turn 2: [get_nearest_airport_by_city, get_flight_cost]  # Merged
Turn 3: set_budget_limit
Turn 4: book_flight  
Turn 5: view_messages_sent
```

### DIFF
--- a/berkeley-function-call-leaderboard/CHANGELOG.md
+++ b/berkeley-function-call-leaderboard/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 All notable changes to the Berkeley Function Calling Leaderboard will be documented in this file.
 
+- [Oct 1, 2025] [#1177](https://github.com/ShishirPatil/gorilla/pull/1177): Fix ground truth for `multi_turn_base_154`.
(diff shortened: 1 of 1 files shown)


HuanzhiMao (approved):
Good catch! Thanks for the fix @chughtapan .

PR #1180, opened 2025-09-15
[BFCL v4] Added Leaderboard Composition BFCL v4

amitojsingh2022:
Added Leaderboard Composition BFCL v4


HuanzhiMao (changes requested):
We should also add a note to the old leaderboard composition table in the v3 blog that it has been outdated, please refer to the new table in xxx.

HuanzhiMao (approved):
LGTM. Thanks @amitojsingh2022 !

PR #1182, opened 2025-09-16
Add Seed-OSS-36B-Instruct model integration

michyli:
Adds support for ByteDance Seed-OSS-36B-Instruct model

### DIFF
--- a/berkeley-function-call-leaderboard/bfcl_eval/model_handler/local_inference/seed_oss.py
+++ b/berkeley-function-call-leaderboard/bfcl_eval/model_handler/local_inference/seed_oss.py
@@ -17,9 +17,9 @@ class SeedOSSHandler(OSSHandler):
     def __init__(self, model_name, temperature, registry_name, is_fc_model, dtype="bfloat16", **kwargs) -> None:
         super().__init__(model_name, temperature, registry_name, is_fc_model, **kwargs)
         self.dtype = dtype
-        # Token IDs from ByteDance-Seed/Seed-OSS tokenizer_config.json:
-        # 2=<seed:eos>, 8=</seed:tool_call>
-        self.stop_token_ids = [2, 8]
+        # Note: Do NOT set stop_token_ids for <seed:tool_call> tags.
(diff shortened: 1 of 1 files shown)


DevonPeroutky:
Was about to make a similar PR until I saw this! Thanks for doing this

berkeley-function-call-leaderboard/bfcl_eval/model_handler/local_inference/seed_oss.py
```diff
@@ -0,0 +1,284 @@
...
+        if function and len(function) > 0:
+            formatted_prompt += "\n\n# Tools\n\nYou may call one or more functions to assist with the user query.\n\n"
+            formatted_prompt += "You are provided with function signatures within <tools></tools> XML tags:\n<tools>\n"
+            for func in function:
+                formatted_prompt += f"{json.dumps(func, indent=2)}\n"
+            formatted_prompt += "</tools>\n\n"
+            
+            formatted_prompt += "For each function call, return the function call in the exact format:\n"
+            formatted_prompt += "[function_name(parameter1=value1, parameter2=value2)]\n\n"
+            formatted_prompt += "IMPORTANT:\n"
```
HuanzhiMao, line 204: Hi @michyli , where did this section come from? For example, the word `IMPORTANT` never shows up in the Jinja2 template you pasted in the doc string. The `_format_prompt` function logic seems very different than the Jinja2 template.

michyli: So the Jinja template is Seed's full native format, and it uses XML-style tool calls and includes functionalities not needed for leaderboard evals, like the think budget. The leaderboard expects function calls in python-style format ([function_name(param=value)]), which is different than the native XML style tool calls Seed has. The `IMPORTANT` section is there to ensure the output format of the model matches what the leaderboard expects, and not default to its native XML format or generate placeholder function name, causing evals to fail. I've had previous runs without this section and the output format doesn't match the expected and fails the evals.

PR #1186, opened 2025-09-19
Fix, openai messsages in BFCL evaluation of multi-turn

hexists:
Hello.

Thank you for providing such a great evaluation tool.

While conducting an evaluation using BFCL api_inference, I discovered an error in the assistant message concatenation during the multi-turn evaluation process, and I fixed it.

Please review this PR.


HuanzhiMao:
Hi @hexists ,
Thanks for the PR! Could you share the steps to reproduce the bug your change addresses so I can verify locally?

hexists:
Hello, @HuanzhiMao .

Would you like to try this command?

```
#Set test_case_ids_to_generate.json
{
"simple": [],
"irrelevance": [],
"parallel": [],
"multiple": [],
"parallel_multiple": [],
"java": [],
"javascript": [],
"live_simple": [],
"live_multiple": [],
"live_parallel": [],
"live_parallel_multiple": [],
"live_irrelevance": [],
"live_relevance": [],
"multi_turn_base": ["multi_turn_base_0"],
"multi_turn_miss_func": [],
"multi_turn_miss_param": [],
"multi_turn_long_context": [],
"multi_turn_composite": []
}

#Run
$ bfcl generate --model gpt-4.1-2025-04-14-FC --run-ids --include-input-log --num-threads 1
```

The `ResponseOutputMessage object` is set to `model_responses_message_for_chat_history`.

HuanzhiMao:
Thanks for the information!

I tried it on my end, and the command seems to be working fine (on the current main branch, without applying your patches). [BFCL_v4_multi_turn_base_result.json](https://github.com/user-attachments/files/22588057/BFCL_v4_multi_turn_base_result.json)

hexists:
Looking at the 'result.json' file alone, there seems to be no problem.
However, if you examine the API call result stored as model_response_data, you can see that the object is still stored.

https://github.com/ShishirPatil/gorilla/blob/main/berkeley-function-call-leaderboard/bfcl_eval/model_handler/api_inference/openai_response.py#L138

https://github.com/ShishirPatil/gorilla/blob/main/berkeley-function-call-leaderboard/bfcl_eval/model_handler/api_inference/openai_response.py#L157
```
[ResponseFunctionToolCall(arguments='{"a":false}', call_id='call_******', name='ls', type='function_call', id='fc_******', status='completed')]
```
A typical message is:
```
[{'type': 'function_call', 'name': 'ls', 'arguments': '{"a":false}', 'call_id': 'call_******', 'status': 'completed'}]
```

HuanzhiMao:
Sorry for the stupid question; I'm a bit confused. Why is it a problem that the object (assume you mean the `ResponseFunctionToolCall`) is stored?

hexists:
@HuanzhiMao 

That's ok.
OpenAI Response API input is guided as follows: https://platform.openai.com/docs/guides/text
```
[
{
"id": "msg_67b73f697ba4819183a15cc17d011509",
"type": "message",
"role": "assistant",
"content": [
{
"type": "output_text",
"text": "Under the moonlight, the unicorn Luna danced across the twinkling stars, leaving a path of dreams for all sleeping children.",
"annotations": []
}
]
}
]
```
Function call processing is guided as follows: https://platform.openai.com/docs/guides/function-calling#handling-function-calls
```
[
{
"id": "fc_12345xyz",
"call_id": "call_12345xyz",
"type": "function_call",
"name": "get_weather",
"arguments": "{\"location\":\"Paris, France\"}"
},
{
"id": "fc_67890abc",
"call_id": "call_67890abc",
"type": "function_call",
"name": "get_weather",
"arguments": "{\"location\":\"Bogota, Colombia\"}"
},
{
"id": "fc_99999def",
"call_id": "call_99999def",
"type": "function_call",
"name": "send_email",
"arguments": "{\"to\":\"bob@email.com\",\"body\":\"Hello, bob\"}"
}
]
```

The model was trained to follow this format, and if you run a custom LLM using the OpenAI-compatible API, chat_template will also assume and process the above input.

High-performance LLMs like OpenAI handle object-oriented input well, but I think it's generally better to process it according to the traditional format.

HuanzhiMao:
@hexists thanks for the detailed follow-up.

I think we might be talking about two different layers:

1. API schema. In the Responses API, tool calls are first-class items in `response.output`, and the official example simply appends `response.output` to the running `input_list` for the next turn — no pre-flattening into a separate “messages” shape. (See https://platform.openai.com/docs/guides/function-calling#function-tool-example)
2. 
```
# Create a running input list we will add to over time
input_list = [
    {"role": "user", "content": "What is my horoscope? I am an Aquarius."}
]

# 2. Prompt the model with tools defined
response = client.responses.create(
    model="gpt-5",
    tools=tools,
    input=input_list,
)

# Save function call outputs for subsequent requests
input_list += response.output
```

3. Local representation/serialization. The ResponseFunctionToolCall you see in model_response_data is just the SDK’s typed wrapper around that JSON. Storing that object is expected; what matters is that, when we make the next request, we provide whatever the target provider expects. OpenAI’s SDK will serialize these objects to the correct wire format.

So I don’t think this is about the LLM’s “performance.” It’s about the API contract. As long as the adapter serializes to the provider’s expected schema at call time, keeping the structured tool-call objects in history data is fine (and matches the OpenAI sample).

hexists:
@HuanzhiMao 

Thank you for the detailed explanation.
As you mentioned, the OpenAI Responses API seems to internally receive and process the messages object as is. I’ve reviewed the `OpenAIResponsesHandler` code and agree to use it in accordance with OpenAI’s guidelines.

Let me explain the LLM performance issue.
First, thank you for providing both the `API-based model` and the `locally hosted OSS model`. While using several handlers (e.g., FunctionalHandler) that inherit from `OpenAICompletionsHandler`, I noticed that messages was being passed directly to the model without any additional processing(flatten). (I checked this through logs when running the model with vllm.) Since this input format may not align with the model’s expected input, it could potentially affect performance. To address this, I’ve submitted a PR.

Please share your feedback.

HuanzhiMao:
This could be an issue on the VLLM side. Could you provide the logs? I can follow up with the VLLM team and get back to you. 

I confirmed with the VLLM team yesterday that the user is expected to interact with the VLLM OpenAI-compatible server in the same way as the official OpenAI endpoint. In other words, the user is supposed to use `input_list += response.output` to add the assistant message to the chat history, and VLLM is capable of handling this.

hexists:
@HuanzhiMao 

To deliver logs, I checked two external models and a custom model.
These are the `meetkai/functionalary-small-v3.1-FC` model and the `Qwen3-32B` model.

I took logs from the `meetkai/functionalary-small-v3.1-FC` model and compared the results. As described above, `input_list += response.output` worked properly. The ChatCompletionMessage object was being flattened and used in bfcl messages.
Functionary handled ChatCompletionMessage internally and Qwen3 handled it in vllm.

The custom model lacked processing logic for ChatCompletionMessage, resulting in the ChatCompletionMessage being used as is. I registered this issue without considering the model's post-processing logic.

Thanks for the great learning. I'll need to add flattening logic for the custom model.
Thank you!

I'll close the PR.

PR #1206, opened 2025-10-04
[BFCL] - Additional Dataset Fixes, Builds off Issues 1133 PR

amitojsingh2022:
multi_turn_base_11: uses find to find all the files and not ls .. could specific prompt to not use find or use the most basic function – separate PR incoming soon

multi_turn_base_15: change last entries to last entry

multi_turn_base_21: Addressed in #1160 

multi_turn_base_23: Addressed in #1160. echo doesn’t have ability to create new file (need touch)

multi_turn_base_24: add the word and then to remove confusion and have the ordering match the ground truth

multi_turn_base_33: remove the part about jotting them down to avoid any confusion and keep the ground truth as correct

multi_turn_base_34: added the word whole to avoid confusion on using the last line or not, keeping the ground truth to be correct

multi_turn_base_39: added the start as Enter the project and populate… this way it knows that cd is required and keeps everything consistent with the possible answers 

Multi_turn_base_93: Updated user query to make the instruction clearer. 

Multi_turn_base_104: Addressed in #1160 

multi_turn_base_117: Removed make_transaction, and replaced it with withdraw_funds. 

multi_turn_base_162: Updated the implementation of `compute_exchange_rate` to always return a float. 

Multi_turn_base_178: Updated the query to include the booking_id information. 

Multi_turn_base_185: just added business CLASS journey to take away any and all confusions that may have resulted in a model to check the pricing for economy class

multi_turn_base_187: Updated the query to include the booking_id information. 

multi_turn_base_194: Updated the travel_booking backend to remove token expiration decrement on invalid access token


chughtapan:
@amitojsingh2022 
1. multi_turn_base_178 and multi_turn_base_187 -- I don't understand your comment here: *booking_record is in the initial config and is provided, the model should have enough information to get the correct ground truth for this prompt.* The initial config is NOT provided in the system instructions or the user prompt, and there is no tool which allows to retrieve bookings.
2. multi_turn_base_194 -- This deviates from the docs - https://gorilla.cs.berkeley.edu/blogs/13_bfcl_v3_multi_turn.html - "taking extra steps should not be penalized", "model result is considered correct if it contains the ground truth as a subset, 
even if it contains additional function calls or takes a different trajectory"
3. multi_turn_117 - get_account_info can provide the account id so this should be valid
4. multi_turn_base_104 -- same point as 194 above -- extra steps should not be penalized
5. multi_turn_base_21 -- "ensure" both are identical is confusing - consider changing the word to something else e.g., "check if"
6. multi_turn_base_93 -- "I would like to increase the amount of fuel in my car to completely full, but first I need to ascertain the present level to determine the appropriate amount to add" -- I think that instructions is unnecessarily complex and ambiguous, and it's hard to understand that you only want to check fuel. Consider rewriting it to say "I would like to check how much fuel can be added or something.."

ShishirPatil (approved):

HuanzhiMao:
> @amitojsingh2022
> 
> 1. multi_turn_base_178 and multi_turn_base_187 -- I don't understand your comment here: _booking_record is in the initial config and is provided, the model should have enough information to get the correct ground truth for this prompt._ The initial config is NOT provided in the system instructions or the user prompt, and there is no tool which allows to retrieve bookings.
> 2. multi_turn_base_194 -- This deviates from the docs - https://gorilla.cs.berkeley.edu/blogs/13_bfcl_v3_multi_turn.html - "taking extra steps should not be penalized", "model result is considered correct if it contains the ground truth as a subset,
>    even if it contains additional function calls or takes a different trajectory"
> 3. multi_turn_117 - get_account_info can provide the account id so this should be valid
> 4. multi_turn_base_104 -- same point as 194 above -- extra steps should not be penalized
> 5. multi_turn_base_21 -- "ensure" both are identical is confusing - consider changing the word to something else e.g., "check if"
> 6. multi_turn_base_93 -- "I would like to increase the amount of fuel in my car to completely full, but first I need to ascertain the present level to determine the appropriate amount to add" -- I think that instructions is unnecessarily complex and ambiguous, and it's hard to understand that you only want to check fuel. Consider rewriting it to say "I would like to check how much fuel can be added or something.."

Hi @chughtapan ,
Thank you for bringing this up. Some of the entries have been addressed back in #1160; the description for this PR is outdated, which results in some confusion. The rest of the entries have been fixed in #1224.

PR #1213, opened 2025-10-14
feat: ✨ support Llama 3.2 Models with chat/completions request

sidhartha-roy:
This PR adds support for evaluating the Llama3.2-1B model with chat/completions requests. This is helpful in evaluating Llama models that only support chat/completions endpoint. For example, with Protopia where a stained glass transform transforms in the input text to obfuscated embeddings and we evaluate the Llama models on the obfuscated prompt. These are the baseline scores.

The scores comparing Llama 3.2 1B with the completions and chat/completions endpoint that is proposed in the PR.
image
image


sidhartha-roy:
@HuanzhiMao, I would really appreciate it if you could take a look at this PR.

sidhartha-roy:
@ShishirPatil can you please take a look at this PR?

PR #1218, opened 2025-10-17
feat: add FC-model consistency check script for model_config.py

zhaoshengbo-source:
The script automatically checks that every ModelConfig entry in model_config.py follows the rule:
If "FC" appears in the key, model_name, or display_name, then is_fc_model must be True. 
Otherwise, is_fc_model should be False or unset. 
If FC appears but  is_fc_model=Flase , it will be reported as error, 
if no FC appears but  is_fc_model=True, it will be reported as a warning.


ShishirPatil:
@zhaoshengbo-source I am wondering if rather than a script, do we just need a warn to catch this? 
Given a model can have FC enabled, but might also be used in prompting mode, just a simple warn at run-time might suffice.

PR #1219, opened 2025-10-18
fix: stabilize tqdm progress bar and prevent log outputs from breaking layout

zhaoshengbo-source:
Pin tqdm bar at bottom with dynamic refresh
Replace all print() with tqdm.write() to prevent bar displacement


ShishirPatil (approved):
Thanks!

PR #1243, opened 2025-11-03
edit the data for mutil turn base

Ericwangjianqi:
Edited some data for muti turn base. The edition for the diff function calling is not tested btw.


ShishirPatil:
Thanks for the PR @Ericwangjianqi  
> The edition for the diff function calling is not tested btw.

Do you mind testing this and updating us when it's done so I can go ahead and merge it!

PR #1263, opened 2025-12-05
Remove pathlib from pyproject.toml dependencies

ShoupingShan:
pathlib has been deprecated after Python 3.7 and carries a naming conflict risk with the standard library.
image


HuanzhiMao (approved):
Thanks for the PR @ShoupingShan !

PR #1269, opened 2025-12-17
[Fix] Correct underscore_to_dot parameter for nanbeige3.5-pro-thinking

fffflust:
**Issue:** `underscore_to_dot` was incorrectly set to `False` (should be `True`)

**Fix:** Set `underscore_to_dot=True` for nanbeige3.5-pro-thinking

**Note:** Only re-evaluation needed, no re-generation required


HuanzhiMao (approved):
LGTM

PR #1290, opened 2026-02-06
fix: :bug: fix issue with unicode that can't be encoded to utf-8

sidhartha-roy:
This issue shows up when the LLM responds with unicode that can't be encoded to utf-8 format in the evaluating the results.


sidhartha-roy:
@HuanzhiMao @ShishirPatil Please review this PR when you get a chance!

PR #1291, opened 2026-02-06
fix: :bug: model path error for evaluating with locally deployed vllm

sidhartha-roy:
Fix issue when the model config is locally available config instead of downloading it from huggingface


sidhartha-roy:
@HuanzhiMao @ShishirPatil Please review this PR when you get a chance!

